PR #2, opened 2020-11-06
[WIP] Introductory tutorial for custom property packages

andrewlee94:
## Proposed changes:
- Adds introductory tutorials for both physical property and reaction packages

### Legal Acknowledgement


TimBartholomew (commented):
I think this start is looking good. These tutorials are much needed.

**Custom Physical Property Packages**
- Should we include the source for specific heat parameters – cell 6
- I have two comments on add_mole_fraction_constraint – cell 10
   - It looks like scaling is hard coded in, shouldn’t we just use the scaling methods instead?
   - I thought this type of formulation leads to degeneracy (at least that was my interpretation from one of Robby’s presentations)
- I think all self should be replaced by b in enth_rule. I think using b is our standard approach, which maybe needs some text explaining why we use b even though it equals self. – cell 12
- Remember to describe why the arguments in get_material_flow_terms and get_enthalpy_flow_terms includes phase, even though none of the variables include a p – cell 13
- Maybe the default_material_balance_type and default_energy_balance_type should be talked about and presented before the get_terms methods. – cell 13

**Custom Reaction Property Packages**
Comments
- Are you planning on writing out the equations so that people can see how that translates to the stoichiometry? I think that would be helpful.
- Should we include the source for the reaction parameters? – cell 5
- Remember to also demonstrate the kinetic reactions.
- I would assert the units right after the creation of the model in cell 13, rather than cell 19

andrewlee94:
@TimBartholomew The descriptive text is coming - there will eventually be text between each of the code cells to describe what is happening, and that will include the sources for the parameters.

* I think scaling is beyond the scope of this tutorial - that is a topic in its own right. I think this tutorial should use hard coded scaling, and only mention the advanced tools.
* `self` is actually more standard than `b`, and in cases like this `self` is probably the better name to use. We are looking at standardizing the code-base to use `self` instead of `b` in most places at some point.
* default balance types are less important than the flow terms, and do different things, so I think they are in the right place for now.
* The reaction example has both an equilibrium and a kinetic reaction in it.

ksbeattie (changes requested):
This should not be merged into the `old_main` branch but the `main` branch.

This can be changed by clicking the "Edit" button and changing the "base" branch for this PR.

PR #5, opened 2020-11-10
Enable CI workflow with GitHub Actions

lbianchi-lbl:
## Summary

This PR adds the first GitHub actions CI workflow running the non-integration tests.

The current version of the workflow seems to be running just fine. That said, it doesn't have feature parity with the CircleCI version yet (e.g. currently runs only one version of Python), and I'd still like to streamline things a bit. However, in the interest of time, maybe it would be better to get this basic version in first, and then do the rest of the changes as a separate PR afterwards.

### Legal Acknowledgement

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,8 +46,8 @@ jobs:
           path: test-reports
           destination: test-reports
       - store_artifacts:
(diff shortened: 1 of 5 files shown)


dangunter (approved):
Looks good

PR #17, opened 2020-12-03
Updating parameter estimation tutorial with new data set

jghouse88:
## Proposed changes:
- Updated the data set for this problem
- Corresponding changes in the testing cells to reflect new values
- Changed initial values for the parameter in the model

### Legal Acknowledgement


dangunter (changes requested):
This is not something you changed, but looking at it again, I think that since the 2 "DMF" example notebooks form a sequence, we should add a link to the second at the bottom of the first. I ran them both and it works pretty well, though. I can't comment on the numeric changes other than the examples still work.

dangunter:
@jghouse88 while you're in there can you correct the typo at the top of `Tutorials/Advanced/ParamEst/DMF_flash_unit_Model_with_NRTL_solution_testing.ipynb` -- somehow an "i=" slipped into the first sentence after the Note.

jghouse88:
@dangunter I have addressed your comments. The links work after they get built, right?

dangunter (approved):
Looks good.

jghouse88:
@andrewlee94 or @ksbeattie can you please take a look at this for the second review?

andrewlee94 (approved):
Looks OK to me.

Issue #35, opened 2021-02-25
Build error with examples-pse 1.9.0rc0 in notebook Examples/SurrMod/ALAMO/ALAMO_six_hump_camel.ipynb


lbianchi-lbl:
Notebook cell where the error occurs:

```py
hasBaron = False
try:
    solver_opt = pyo.SolverFactory('baron')
    hasBaron = solver_opt.available()
except ApplicationError as err:
    print("Partial test completed. %s"%err)


if alamo_ran and hasBaron:
    model = pyo.ConcreteModel()
    opt = pyo.SolverFactory('baron')
    model.x1 = pyo.Var()
    model.x2 = pyo.Var()
    def pyomo_model(model):
        import z1
        return z1.f(model.x1,model.x2)
    model.obj = pyo.Objective(rule = pyomo_model)
    results = opt.solve(model)
    model.solutions.store_to(results)
    print(results)
```

Traceback:

```txt
ValueError                                Traceback (most recent call last)
<ipython-input-1-b0dff88c0fa0> in <module>
     17     model.obj = pyo.Objective(rule = pyomo_model)
     18     results = opt.solve(model)
---> 19     model.solutions.store_to(results)
     20     print(results)

~/anaconda3/envs/examples-rel/lib/python3.7/site-packages/pyomo/core/base/PyomoModel.py in store_to(self, results, cuid, skip_stale_vars)
    322                 else:
    323                     vals = vals[1]
--> 324                 vals['Value'] = value(obj)
    325                 soln.variable[ sm.getSymbol(obj, labeler) ] = vals
    326             entry = soln_._entry['constraint']

pyomo/core/expr/numvalue.pyx in pyomo.core.expr.numvalue.value()

pyomo/core/expr/numvalue.pyx in pyomo.core.expr.numvalue.value()

ValueError: No value for uninitialized NumericValue object x1
ValueError: No value for uninitialized NumericValue object x1
```

mengleCMU:
PR in response to thise issue: https://github.com/IDAES/examples-pse/pull/36

PR #36, opened 2021-02-25
Termination condition check for the six_hump_camel output

mengleCMU:
## Fixes # .
Fixing error when storing results by checking the solver first.

## Proposed changes:
-

### Legal Acknowledgement


lbianchi-lbl (approved):
The notebook works without issue with the changes in this PR.

I've tried a few times, but unfortunately I haven't been able to reproduce the original error described in #35 using the version prior to these changes, so I don't have true "before/after" evidence of the bug being fixed.

All in all, since this version is working and has some additional check against errors, I'd say we can merge this in and go back to this if the error in #35 rears its head again in the future.

blnicho (approved):

PR #42, opened 2021-04-15
Address errors in build script on notebook timeout

### LINKED ISSUE #41: Notebook build fails with "Failed due to error: 'ParallelNotebookWorker' object has no attribute 's'" on timeout
I believe this is caused when catching the `TimeoutException` [here](https://github.com/IDAES/examples-pse/blob/722dd5037ca0c8087e03141cbc552edd75d1fd0b/build.py#L975-L977): trying to access `self.s` results in an `AttributeError` since `ParallelNotebookWorker` does not set this attribute, ~possibly because it doesn't inherit from the `Builder` base class~ **EDIT** (`ParallelNotebookWorker` is used internally by `NotebookBuilder`, so the inheritance remark probably doesn't apply).

This occurred when running notebooks as part of the integration checks for IDAES/idaes-pse#249, but I think could happen whenever a `TimeoutError` is thrown in a `ParallelNotebookWorker`, which I assume happens when a notebook's runtime exceeds a certain threshold.

I could work on a quick fix once @dangunter chimes in on what would be the best way for `ParallelNotebookWorker` instances to set/initialize `s`.

lbianchi-lbl:
## Proposed changes:
- Address error in #41 (proposed solution)
- Address another possible error detected by pylint
- Add basic pylint check to help detect future errors in build code
- (Unrelated) Fix property key name in testing cell of `src/Tutorials/Advanced/ParamEst/parameter_estimation_NRTL_using_state_block_solution_testing.ipynb` that would cause `KeyError` when running notebook as-is (i.e. without stripping test cells)

### Legal Acknowledgement


jghouse88 (approved):
LGTM (particularly on the notebook change). I will defer to @dangunter for the rest of the changes.

dangunter (approved):
Looks good

PR #43, opened 2021-04-15
Removing bootstrap from parmest notebooks

jghouse88:
## Fixes # .

https://github.com/IDAES/idaes-pse/issues/293 

## Proposed changes:
- Removing bootstrap cell in the parmest notebooks as this could be the reason for unpredictable convergence problems and long run times. 

### Legal Acknowledgement


jghouse88:
@lbianchi-lbl This might address the long run time issues (maybe) but either way I think it is a good idea to remove the bootstrap cell from the notebooks. I have left it as commented code for users to try on their own.

andrewlee94 (approved):
Looks OK to me.

dangunter (approved):
Looks good

PR #52, opened 2021-07-22
Updated deprecated class, phase, unit definitions in methanol_param_VLE.py

bpaul4:
## Fixes # .
- Fixed state block class definition to remove deprecation warning
- Replaced phase and component definitions with new object forms
- Updated unit lists to pyomo.environ.units (as pyunit objects)

## Proposed changes:
- Line 75, in def build(self) updated‘self.state_block_class’ to ‘self._state_block_class’ to note private class
- Line 79, same def updated phase and component definitions from list to objects (and updated imports as well)
- Line 155, updated default unit definitions to pyunits (and updated imports as well); this item needs review, pyunits.kgmol tag was not recognized and pyunits.kmol results in AssertionError from compressor work value in Custom Compressor notebook.

### Legal Acknowledgement


bpaul4:
Hi Jaffer and Michael, I have been updating the methanol properties package for the Custom Compressor/Heater example notebooks and would appreciate your review. The original file displayed a set of deprecation warnings, my commit fixes these issues but yields a new AssertionError issue in the notebooks. Thanks!

bpaul4:
@jghouse88 When you have a moment, I'd appreciate your insight into the property and state block package files. I believe I missed some unit definitions switching from kgmol (deprecated) to kmol (new Pyomo tag). The two files have commented-out lines that specify the units by brute force using the `units` calls, in case that's part of what is missing, and the temperature is still scaled by a factor of 100. Thanks!

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -106,7 +100,8 @@ def build(self):
         self.gas_constant = Param(within=NonNegativeReals,
                                   mutable=False,
                                   default=0.008314,
-                                  doc='Gas Constant [MJ/(kgmol.K)]')
+#                                  units=pyunits.MJ/pyunits.kmol/pyunits.K,
+                                  doc='Gas Constant [MJ/(kmol.K)]')
```
andrewlee94, line 104: Note that this should be replaced by the value in the IDAES constants module: `from idaes.core.util.constants import Constants`. It will need to have a unit conversion applied to it wherever it is used however.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -127,35 +122,37 @@ def build(self):
 
         self.gamma = Param(within=NonNegativeReals, mutable=True, default=gamma, doc='Ratio of Cp to Cv')
 
-        self.Cp = Param(within=NonNegativeReals, mutable=True, default=Cp, doc='Constant pressure heat capacity [MJ/(kgmol K)]')
+        self.Cp = Param(within=NonNegativeReals, mutable=True, default=Cp,
+#                        units=pyunits.MJ/pyunits.kmol/pyunits.K,
```
andrewlee94, line 126: Units should definitely be specified here.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -127,35 +122,37 @@ def build(self):
...
+                        doc='Constant pressure heat capacity [MJ/(kmol K)]')
 
     @classmethod
     def define_metadata(cls, obj):
         """Define properties supported and units."""
         obj.add_properties(
-            {'flow_mol': {'method': None, 'units': 'kgmol/s'},
+            {'flow_mol': {'method': None, 'units': 'kmol/s'},
```
andrewlee94, line 133: The "units" section of the metadata is not really needed anymore, as units are fully defined by the units metadata below. You could just remove this from all entries.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -127,35 +122,37 @@ def build(self):
...
              'mole_frac': {'method': None, 'units': 'no unit'},
              'temperature': {'method': None, 'units': '100K'},
```
andrewlee94, line 135: This is going to be an issue. The current implementation of units does not allow a lot of freedom here, and everything needs to be based on the units defined in the units metadata.

So, using 100*K (centaKelvins?) is possible, it means that any unit of temperature will be in 100*K. Similarly, it is possible to set thing up so that pressure is in MPa, but it means that you have to use mass/length/time units that work out that way and use them everywhere else.

bpaul4: @jghouse88 explained that the scaling originated from the source example. Removing the factor of 100 in the enthalpy balance (state block file near the bottom) seemed to make the compressor work error worse so I kept it for now.

andrewlee94: The problem here is that scaling should be kept separate from units. The fact that the model has solver issues implies that the scaling tools should be applied, rather than that the units should be changed.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
bpaul4: Would this need to be updated in the state block file as well?

andrewlee94: Yes - in fact there should be no need for this in the parameter block. You should just use it where needed in the state block.

bpaul4: Makes sense, thank you @andrewlee94

bpaul4:
I have been looking over the files per @andrewlee94's suggestions (thank you) and managed to obtain consistent units in the newly committed file versions. However, I am still having assertion errors in the custom compressor and heater notebooks. The compressor value is correct but a factor of 1E5 too large (probably a scaling issue), and the heater value is incorrect. I'm not sure what is causing the issue, assuming the Antoine coefficients are correct I believe the pressure, temperature and energy unit conversions are correctly handled in the property packages.

andrewlee94:
@bpaul4 The value magnitude error you mention sounds more like units than scaling - scaling tends to give fully incorrect values, or a failure. If the magnitude is wrong, that tends to mean there is a missing unit conversion somewhere. Also, Antoine coefficients tend to cause things to break badly as it is an exponential relationship.

Also, it looks like you need to update some of the tests - there appears to be a test for scaled temperature that needs to be updated to checking temperature directly.

bpaul4:
The phase/component definitions in the parameter property package, Pyomo unit definitions throughout both property package files and in-line comments in the notebooks have been updated. The updates resolve the deprecation warnings and assertion errors, and the files should be ready to merge.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -62,37 +63,37 @@ class PhysicalParameterData(PhysicalParameterBlock):
...
+        
+        self.CH4 = Component()
+        self.CO = Component()
+        self.H2 = Component()
+        self.CH3OH = Component()
+        
         # List of components in each phase (optional)
-        self.phase_comp = {"Liq": self.component_list,
-                           "Vap": self.component_list}
+#        self.phase_comp = {"Liq": self.component_list,
```
andrewlee94, line 95: Remove commented code.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -119,43 +114,37 @@ def build(self):
                                      ('H2', 'C'): 3.19,
                                      ('CH3OH', 'A'): 18.5875,
                                      ('CH3OH', 'B'): 3626.55,
-                                     ('CH3OH', 'C'): -34.29}
+                                     ('CH3OH', 'C'): -34.29} # these Antoine coefficients assume mmHG and K
 
         Cp = self.config.Cp
-        Cv = Cp - self.gas_constant.value
+        Cv = Cp - Constants.gas_constant/1000/pyunits.mol*pyunits.kmol # convert gas constant from J/mol-K to MJ/kmol-K
```
andrewlee94, line 120: You should use `pyunits.convert()` here, rather than hard coding the conversion.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -119,43 +114,37 @@ def build(self):
...
         gamma = Cp / Cv
 
         self.gamma = Param(within=NonNegativeReals, mutable=True, default=gamma, doc='Ratio of Cp to Cv')
 
-        self.Cp = Param(within=NonNegativeReals, mutable=True, default=Cp, doc='Constant pressure heat capacity [MJ/(kgmol K)]')
+        self.Cp = Param(within=NonNegativeReals, mutable=True, default=Cp,
```
andrewlee94, line 125: Check the PEP8 standard formatting here - I think that if you need to break arguments over multiple lines then the standard is to put each argument on a new line.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
```diff
@@ -29,13 +29,14 @@
...
                         EnergyBalanceType)
+from idaes.core.util.constants import Constants
 from idaes.core.util.initialization import (fix_state_vars,
                                             revert_state_vars,
                                             solve_indexed_blocks)
 from idaes.core.util.exceptions import ConfigurationError
 
 # Some more inforation about this module
-__author__ = "Jaffer Ghouse"
+__author__ = "Jaffer Ghouse", "Brandon Paul" # updated units by Brandon Paul
```
andrewlee94, line 39: As above.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
```diff
@@ -299,7 +306,9 @@ def rule_comp_mass_balance(self, i):
     def _make_flash_eq(self):
         self.vapor_pressure = Var(self._params.component_list,
                                   initialize=0.101325,
-                                  doc="vapor pressure ",
+                                  domain=NonNegativeReals,
+                                  units=pyunits.MPa, 
+                                  doc="vapor pressure [MPa]",
```
andrewlee94, line 311: Typo: doc string should start with a capital. I also tend not to include units in the doc string when they are explicitly defined.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
```diff
@@ -332,23 +341,25 @@ def rule_Keq(self, i):
                 self.vapor_pressure[i] * self.mole_frac_phase_comp['Liq', i]
         self.eq_Keq = Constraint(self._params.component_list, rule=rule_Keq)
 
-        def rule_P_vap(self, j):
+        def rule_P_vap(self, j): # coefficient converts MPa to mmHg
             return log(7500.6168 * self.vapor_pressure[j]) == \
                 self._params.vapor_pressure_coeff[j, 'A'] - \
                 (self._params.vapor_pressure_coeff[j, 'B'] /
-                 (100*self.temperature - self._params.vapor_pressure_coeff[j, 'C']))
+                (self.temperature*100/pyunits.hectokelvin*pyunits.kelvin - self._params.vapor_pressure_coeff[j, 'C'])) # convert T to K
```
andrewlee94, line 348: Again, you should use `pyunits.convert()` here.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
```diff
@@ -332,23 +341,25 @@ def rule_Keq(self, i):
...
+                               units=pyunits.kmol/pyunits.m**3,
+                               doc="Molar density [kmol/m^3]")
 
         def density_mol_calculation(self, p):
             if p == "Vap":
-                return 10.0 * self.pressure == (self.density_mol[p] *
-                                         self._params.gas_constant *
-                                         self.temperature)
+                return self.pressure == (self.density_mol[p] *
+                                         Constants.gas_constant/1000/pyunits.mol*pyunits.kmol *
```
andrewlee94, line 359: As above.

andrewlee94 (changes requested):
A few comments, the biggest one being to use `pyunits.convert` rather than hard coding conversions. Note that Pyomo should support `hK` for hectoKelvins.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -62,37 +63,37 @@ class PhysicalParameterData(PhysicalParameterBlock):
     CONFIG.declare("Cp", ConfigValue(
         default=0.035,
         domain=float,
```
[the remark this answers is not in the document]
bpaul4: At the moment, I believe the only valid parameters for ConfigValue() are default, domain, description, doc and visibility. The code on line 125 adds Cp as a parameter along with attributes including units.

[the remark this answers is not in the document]
andrewlee94: Sorry, I missed that it was a config argument. In that case I am slightly surprised that it is a config argument.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
bpaul4: I'd like the `doc` tags on lines 254, 258, 311, and 354 to be consistent. Would you recommend removing the units from all of them, as well as line 127 in the the main param_VLE file?

andrewlee94: Doc strings should be consistent. Others might disagree with not including units in the doc string however - I personally prefer not including them to avoid any chance of a mismatch between the docs and the defined units.

bpaul4:
Thank you @andrewlee94 for your suggestions. The files have been updated and all errors have been resolved.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -103,6 +106,7 @@ def build(self):
              3: ["H2", ("Vap", "Liq")],
              4: ["CH3OH", ("Vap", "Liq")]}
 
+# Antoine coefficients assume pressure in mmHG and temperature in K
```
andrewlee94, line 109: This should be indented. It is a good idea to activate a linter in your IDE if you can, as this will highlight these types of issues for you.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -114,17 +118,23 @@ def build(self):
                                      ('H2', 'C'): 3.19,
                                      ('CH3OH', 'A'): 18.5875,
                                      ('CH3OH', 'B'): 3626.55,
-                                     ('CH3OH', 'C'): -34.29} # these Antoine coefficients assume mmHG and K
+                                     ('CH3OH', 'C'): -34.29}
 
         Cp = self.config.Cp
-        Cv = Cp - Constants.gas_constant/1000/pyunits.mol*pyunits.kmol # convert gas constant from J/mol-K to MJ/kmol-K
+        Cv = Cp - pyunits.convert(Constants.gas_constant,
```
andrewlee94, line 124: It isn't really necessary here, but I think it would be good practice to use the Pyomo `value()` method to evaluate this expression. In this case you get by without it as all parts of the expression are constants, but if anything was a variable (if the general sense), then you would see an exception when using this.

src/Examples/Advanced/CustomUnitModels/methanol_state_block_VLE.py
```diff
@@ -345,21 +350,23 @@ def rule_P_vap(self, j): # coefficient converts MPa to mmHg
...
             if p == "Vap":
                 return self.pressure == (self.density_mol[p] *
-                                         Constants.gas_constant/1000/pyunits.mol*pyunits.kmol *
-                                         self.temperature*100/pyunits.hectokelvin*pyunits.kelvin) # convert gas constant to MJ/kmol-K and T to K
+                                         pyunits.convert(Constants.gas_constant,
+                                           pyunits.MJ/pyunits.kmol/pyunits.K) *
+                                         pyunits.convert(self.temperature, pyunits.K))
             elif p == "Liq":
-                return self.density_mol[p] == 11.1  # kmol/m3 # dummy value
+                return self.density_mol[p] == 11.1  # dummy value
```
andrewlee94, line 369: You should apply the units to the constant itself for unit consistency. I.e. `11.1*pyunits.kmol/pyunits.m**3`

andrewlee94 (changes requested):
A few more comments on some things I saw this time.

Also, does the Jupyter notebook contain any test cells? We need to make sure it does so that we know if anything breaks in this example. We should make sure one of these tests included an assertion of unit consistency as well, especially due to the unusually units being used here.

bpaul4:
Thank you @andrewlee94. The two notebooks already have test cells to assert units as well as the model results (work for the compressor, outlet temperature for the heater).

bpaul4:
To clarify, at the end of each notebook there are two test cells:
`from pyomo.util.check_units import assert_units_consistent

assert_units_consistent(m)`

`import pytest
from pyomo.environ import TerminationCondition, value

assert res.solver.termination_condition == TerminationCondition.optimal
assert value(m.fs.compressor.work) == pytest.approx(5.2616, abs=1e-2)`

andrewlee94:
@bpaul4 OK, then it looks like we have a unit consistency check already.

bpaul4:
@andrewlee94 I believe the files match the expected standards for the codebase, and I've addressed all linting/formatting issues mentioned above.

andrewlee94 (approved):
Looks OK to me.

MAZamarripa (approved):
lgtm

Issue #53, opened 2021-07-28
MatOpt example jupyter notebooks' output error


xiangyuy:
Hi, we have noticed that there are errors in the outputs of MatOpt example notebooks. For example, in the output below In[11] in https://idaes.github.io/examples-pse/latest/Examples/MatOpt/monometallic_nanocluster_design_doc.html, it says that the size of the model exceeds the community version's CPLEX' limitation. However, in the source file https://github.com/IDAES/examples-pse/blob/main/src/Examples/MatOpt/monometallic_nanocluster_design.ipynb, the same version of community CPLEX solves the problem successfully and shows no error in the output. Does the website-generation mechanism ignores the original output and tries to generate new outputs on the fly? If so, is it possible that previous version notebooks were used? If the issue arises from CPLEX further reducing the size limitation, we could modify the source files to reduce the size of the examples further or to use NEOS-CPLEX (if the website-generation mechanism supports internet access). Thank you.

ksbeattie:
We do re-run all the notebooks when we cut a release.  We do this because not all notebooks are checked-in with results in them and to ensure things are working correctly, which clearly we are missing in this case.

One thing I noticed in just visually comparing the two notebooks pointed to in the issue is that they look like different problem sizes.  I'm not that familiar with CPLEX but the published one has "Variables : 1920" and the checked-in one has "Variables : 426".  Another difference is that XYs looks to have been run on Windows whereas we run our on Linux - hopefully that doesn't matter, but I suppose it could.

I'll dig more to see if I can find when and how the difference in problem size got introduced.  Let me know if you have any ideas on it.

lbianchi-lbl:
I looked into the history of the notebook and I found a commit which seems to be related to the increase in problem size (as reported by the `Variables: ` line in the CPLEX output: https://github.com/IDAES/examples-pse/commit/e5e27bb80eb127a8973b9d27d2fc60d21bb54995#diff-b36be862be9825ffc2a9975f11846b599a5aca48fb991afce84ef3041af515df

lbianchi-lbl:
@xiangyuy: @ksbeattie and I are working on a way to address this issue. As you correctly assumed, the HTML version of the examples documentation (i.e. the one that's accessible at `https://idaes.github.io/examples-pse`) is generated each time a release of `examples-pse` is cut, by (simplifying) running each notebook using `nbconvert` and then converting the re-run notebook to an HTML page using Sphinx.

[...]

PR #54, opened 2021-07-29
Build configuration to address #53

lbianchi-lbl:
# DO NOT MERGE

This PR is being used to facilitate the steps needed to address #53 in the short term, and is not supposed to be merged into `main` as-is.

## Instructions

Check out the code in this PR:

```sh
mkdir idaes-examples-pr-54 && cd idaes-examples-pr-54

git clone https://github.com/IDAES/examples-pse && cd examples-pse
git fetch origin pull/54/head:pr-54
git checkout pr-54
```

If not already present, install pandoc for your system using the instructions here: (https://pandoc.org/installing.html).

> **NOTE**: pip cannot be used to install pandoc. The `pandoc` package on PyPI is not the pandoc that is needed for the `examples-pse` build; rather, it provides Python bindings to pandoc, and requires pandoc to be installed separately

From an environment where IDAES is installed and fully enabled:

```sh
pip install -r requirements.txt
python build.py --config build.yml -cdr -v -w 1
```

There will be some warnings due to the missing notebooks that have been removed from the build config file just for this PR; they can be ignored.

Once the build succeeds, you can use a web browser to verify that the HTML output is OK by opening the `index.html` file in the desired directory, and then browsing as normal to the remaining pages, e.g.:

```sh
firefox docs/_build/html/Examples/MatOpt/index.html
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,6 +108,7 @@ jobs:
     - name: Check for undetected errors in HTML output
       shell: bash
       run: |
+        grep "WARNING: Could not locate" -r ./docs_test/_build/html/ ; test $? -eq 1
         grep "size limits exceeded" -r ./docs_test/_build/html/ ; test $? -eq 1
     - name: Check for broken links in HTML docs
       shell: bash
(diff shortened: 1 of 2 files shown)


lbianchi-lbl:
This is not needed anymore after merging #55.

PR #62, opened 2021-08-30
Move HDA_flowsheet_with_distillation under Tutorials to address build errors

### LINKED ISSUE #59: Build error for notebook Examples/Flowsheets/HDA_flowsheet_with_distillation_solution_testing.ipynb
The following error occurred when building the full set of examples in preparation for the 1.11.0rc0 release:

```txt
2021-08-25 20:10:56,659 INFO - [Worker 0] Convert notebook name=/home/lbianchi/idaes/examples-pse/src/Examples/Flowsheets/HDA_flowsheet_with_distillation_solution_testing.ipynb: begin
2021-08-25 20:10:56,659 INFO - [Worker 0] Converting: HDA_flowsheet_with_distillation_solution_testing.ipynb
2021-08-25 20:10:56,906 INFO - [Worker 0] Stripped tags from: HDA_flowsheet_with_distillation_solution_testing.ipynb
2021-08-25 20:10:56,906 INFO - [Worker 0] Running notebook: HDA_flowsheet_with_distillation_solution.ipynb
2021-08-25 20:10:58,679 ERROR - [Worker 0] Notebook execution failed: execution error for '/tmp/tmpzahzi5y0/HDA_flowsheet_with_distillation_solution.ipynb': An error occurred while executing the following cell:
------------------
import hda_reaction_kinetic as reaction_props
from idaes.generic_models.properties.activity_coeff_models.\
    BTX_activity_coeff_VLE import BTXParameterBlock

# Add the folder containing the thermodynamic package to the system path
import sys 
sys.path.insert(0, '../..')

from Tutorials.Basics.hda_ideal_VLE import HDAParameterBlock
------------------

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/tmp/ipykernel_20539/3355508218.py in <module>
      7 sys.path.insert(0, '../..')
      8 
----> 9 from Tutorials.Basics.hda_ideal_VLE import HDAParameterBlock

ModuleNotFoundError: No module named 'Tutorials'
ModuleNotFoundError: No module named 'Tutorials'

2021-08-25 20:10:58,679 ERROR - [Worker 0] Execution failed: generating partial output for '/home/lbianchi/idaes/examples-pse/src/Examples/Flowsheets/HDA_flowsheet_with_distillation_solution_testing.ipynb'
2021-08-25 20:10:58,679 INFO - [Worker 0] Convert notebook name=/home/lbianchi/idaes/examples-pse/src/Examples/Flowsheets/HDA_flowsheet_with_distillation_solution_testing.ipynb: end, ok=False duration=2.0s
```

The likely cause for this is the `sys.path.insert(0, '../../')` used to import packages into the notebook. Manipulating `sys.path` in this fashion should be avoided in general, and in particular when used for IDAES examples notebooks since it's almost guaranteed not to work.

lbianchi-lbl:
## Motivation

- While I was looking at the `HDA_flowsheet_with_distillation` notebook to investigate #59, I realized that one quick way to address the build errors would be to move it from `Examples/Flowsheets` to `Tutorials/Basics`
- This addresses the build errors by removing the need for the fragile workaround involving `sys.path`, as the required file is now in the same directory as the notebook and can be imported directly
- Regardless of the build errors, this looks like a reasonable solution since the notebook is written as a tutorial anyway

## For reviewers

- Main question: **does the new location in `Tutorials` make sense for this particular notebook?**
- This solution seems to address the issues without requiring any change to the build machinery, so I think we should consider it anyway for the Aug release independently from a more permanent solution

## Proposed changes:
- Move `HDA_flowsheet_with_distillation.ipynb` + additional material from `scr/Examples/Flowsheets` to `src/Tutorial/Basic`
- Remove the `sys.path` workaround to import the Python file into the notebook since it's not needed anymore
- Remove the (now empty) `Examples/Flowsheets` directory from `src/` and `docs/`

### Legal Acknowledgement


andrewlee94:
@lbianchi-lbl Whilst this is written as a tutorial, and we intend it to become one eventually, it is not ready to replace the existing tutorials yet. Hence, we put it in examples for now so as not to confuse users with multiple tutorials with similar names.

lbianchi-lbl:
@andrewlee94 sounds good, thanks for the clarification. I'll close this PR and open another one with the complementary solution to this one.

PR #63, opened 2021-08-30
Address build errors in HDA_flowsheet_with_distillation by copying required property package to notebook directory

lbianchi-lbl:
## Motivation

- Build error in `HDA_flowsheet_with_distillation` notebook is caused by trying to import a .py file located outside of the notebook's working directory, which is not supported by the build machinery
- Having a copy of the required property package in the notebook's directory addressed this
- This creates duplication in the two copies of the same `hda_ideal_VLE.py` file which can be considered a tradeoff for the simplicity of this solution (as opposed to modifying the build machinery to support this single case)

## Proposed changes:
- Copy required `hda_ideal_VLE.py` property package file to the notebook's directory
- Add `Examples/Flowsheets` directory to CI build config file

### Legal Acknowledgement

### DIFF
--- a/build-ci.yml
+++ b/build-ci.yml
@@ -24,6 +24,7 @@ notebook:
     directories:
       - source: Examples/Advanced/CustomUnitModels
       - source: Examples/Advanced/CustomProperties
+      - source: Examples/Flowsheets
 #      - source: Examples/Advanced/DataRecon
       - source: Examples/MatOpt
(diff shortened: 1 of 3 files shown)


andrewlee94 (approved):
Looks good to me.

radhakrishnatg (approved):
@lbianchi-lbl Thank you for making the changes! They look good to me.

PR #65, opened 2021-09-22
Fixing use of .value on expressions

andrewlee94:
## Fixes None.

A recent change in IDAES revealed that the ASU notebook was using `.value` to retrieve values from `VarLikeExpressions` which is no longer supported.

## Proposed changes:
- Change to use `value()` method when checking model outputs in ASU notebook

### Legal Acknowledgement


eslickj (approved):

jsiirola (approved):

jghouse88:
Only concern is the restoration at the last step.

PR #67, opened 2021-10-06
Adding Reactor Examples to IDAES

bpaul4:
## Proposed changes:
- Example notebooks with supporting property/reaction package files (CSTR currently, more later)
- Packages demonstrate integrating Generic Property Package and Generic Reaction Package Frameworks for a multi-component liquid system with a kinetics-based rate reaction into IDAES flowsheet optimization
- Updated Perrys method to handle an additional liquid density equation form, per https://github.com/IDAES/idaes-pse/pull/543

### Legal Acknowledgement


bpaul4:
@andrewlee94 @jghouse88 all formatting, docs and pytest issues have been addressed, and the changes are ready for review. Thanks!

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "## Importing required pyomo and idaes components\n",
+    "\n",
+    "\n",
+    "To construct a flowsheet, we will need several components from the pyomo and idaes package. Let us first import the following components from Pyomo:\n",
+    "- Constraint (to write constraints)\n",
+    "- Var (to declare variables)\n",
+    "- ConcreteModel (to create the concrete model object)\n",
+    "- Expression (to evaluate values as a function of variables defined in the model)\n",
+    "- Objective (to define an objective function for optimization)\n",
+    "- SolverFactory (to solve the problem)\n",
```
andrewlee94, line 56: Is this needed? I notice that you import `get_solver` later on, and we tend to use that as the default approach in most of our code.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Importing required pyomo and idaes components\n",
+    "\n",
+    "\n",
+    "To construct a flowsheet, we will need several components from the pyomo and idaes package. Let us first import the following components from Pyomo:\n",
```
andrewlee94, line 50: I think Pyomo and IDAES should be capitalised in this context.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "from idaes.core import FlowsheetBlock\n",
+    "from idaes.generic_models.properties.core.generic.generic_property import (\n",
+    "        GenericParameterBlock)\n",
+    "from idaes.generic_models.properties.core.generic.generic_reaction import (\n",
+    "        GenericReactionParameterBlock)\n",
+    "from idaes.generic_models.unit_models import (Mixer,\n",
+    "                                              Heater,\n",
+    "                                              CSTR)\n",
+    "\n",
+    "from idaes.core.util.constants import Constants as c\n",
```
andrewlee94, line 97: Minor comment, but I would use a longer name than `c` to help remind the user what this is when it gets used later on. Also, I cannot recall if this was actually used in the example, so check to confirm if it is needed.

I would tend to do `from idaes.core.util.constants import Constants` rathe than adding an alias (unless the name `Constants` is going to cause a conflict.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "ParamEst parameter estimation: https://idaes-pse.readthedocs.io/en/stable/user_guide/workflow/data_rec_parmest.html?highlight=paramest  \n",
+    "HELMET thermodynamic estimation: https://idaes-pse.readthedocs.io/en/stable/user_guide/modeling_extensions/surrogate/helmet/index.html  \n",
+    "RIPE reaction estimation: https://idaes-pse.readthedocs.io/en/stable/user_guide/modeling_extensions/surrogate/ripe/index.html  \n",
+    "\n",
+    "Let us import the following modules from the same directory as this jupyter notebook:\n",
+    "      <ul>\n",
+    "         <li>egprod_ideal as thermo_props</li>\n",
+    "         <li>egprod_reaction as reaction_props </li>\n",
+    "      </ul>\n",
+    "</div>"
```
andrewlee94, line 139: Not sure what this is, and it might just be GitHub, but this appeared in the rendering of the notebook.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "Let us first add an Expression to convert the product flow from mol/s to MM lb/year of ethylene glycol. We see that our molecular weight exists in the thermo property package, so we may use that value for our calculations."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.eg_prod = Expression(expr=m.fs.R101.outlet.flow_mol_phase_comp[0, \"Liq\", \"ethylene_glycol\"]*3600*24*365 # converting /s to /year\n",
```
andrewlee94, line 294: We should use the Pyomo unit conversion tools here to show how it is done properly.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+      "15\n",
+      "0\n",
+      "1\n",
+      "2\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(degrees_of_freedom(m))  # (6 per stream x 2 streams) + (3 from units) = 15 DOF\n",
+    "print(degrees_of_freedom(m.fs.M101) - 12)  # 2 inlets, so subtract those 12 DOF\n",
```
andrewlee94, line 346: In the spirit of a Diataxis tutorial, I would be inclined to just tell the user how many degrees of freedom there are at this point, and what we will fix to address that,

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "We will now be fixing the feed stream to the conditions shown in the flowsheet above. As mentioned in other tutorials, the IDAES framework expects a time index value for every referenced internal stream or unit variable, even in steady-state systems with a single time point $ t = 0 $. The non-present components in each stream are assigned a very small non-zero value to help with convergence and initializing. Based on stoichiometric ratios for the reaction, 80% conversion and 200 MM lb/year (46.4 mol/s) of ethylene glycol, we will initialize our simulation with the following calculated values:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.M101.reagent_feed.flow_mol_phase_comp[0, \"Liq\", \"ethylene_oxide\"].fix(58.0)\n",
```
andrewlee94, line 382: I believe we can no do the following:

```
m.fs.M101.reagent_feed.flow_mol_phase_comp[0, "Liq", "ethylene_oxide"].fix(58.0*pyunits.mol/pyunits.s)
```

I.e. we can include units when fixing/setting variable values. If this works, we should begin doing it throughout our examples to properly show units and the formal way to do things.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "We will assume that 80% of our total reactor volume is working volume (CSTR reaction volume). For the CSTR, we have to define the conversion in terms of ethylene oxide as well as the CSTR reaction volume. This requires us to create new variables and constraints relating reactor properties to stream properties.  Note that the CSTR reaction volume variable (m.fs.R101.volume) does not need to be defined here since it is internally defined by the CSTR model. Additionally, the heat duty is not fixed, since the heat of reaction depends on the reactor conversion (through the extent of reaction and heat of reaction)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.R101.total_volume = Var(initialize=6.9225, bounds=(0, 15))  # m^3\n",
```
andrewlee94, line 432: We should be declaring units on all these variables.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "m.fs.R101.conversion = Var(initialize=0.80, bounds=(0, 1))  # fraction\n",
```
andrewlee94, line 433: Even here, we should set `units=pyunits.dimensionless`

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "\n",
+    "m.fs.R101.conv_constraint = Constraint(\n",
+    "    expr=m.fs.R101.conversion*m.fs.R101.inlet.\n",
+    "    flow_mol_phase_comp[0, \"Liq\", \"ethylene_oxide\"] ==\n",
+    "    (m.fs.R101.inlet.flow_mol_phase_comp[0, \"Liq\", \"ethylene_oxide\"] -\n",
+    "     m.fs.R101.outlet.flow_mol_phase_comp[0, \"Liq\", \"ethylene_oxide\"]))\n",
+    "\n",
+    "m.fs.R101.vol_constraint = Constraint(\n",
```
andrewlee94, line 441: A comment on why 80% is used here might be useful (there might be one that I missed).

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "Now, we need to add the design constraints and unfix the decision variables as we had solved a square problem (degrees of freedom = 0) until now, as well as set bounds for the design variables:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.eg_prod_con = Constraint(expr=m.fs.eg_prod >= 200)  # MM lb/year\n",
```
andrewlee94, line 801: We should be adding units to these constants too.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1123 @@
...
+    "        GenericReactionParameterBlock)\n",
+    "from idaes.generic_models.unit_models import (Mixer,\n",
+    "                                              Heater,\n",
+    "                                              CSTR)\n",
+    "\n",
+    "from idaes.core.util.constants import Constants as c\n",
+    "from idaes.core.util import get_solver\n",
+    "from idaes.core.util.model_statistics import degrees_of_freedom\n",
+    "from idaes.core.util.initialization import (propagate_state,\n",
+    "                                            fix_state_vars,\n",
```
andrewlee94, line 101: I did not see `fix_state_vars`, `revert_state_vars` or the logger being used here.

andrewlee94 (changes requested):
I mostly focused on the code for now and had a number of comments.

The main one is that we should be very formal about defining and using units as this is an example; we need to show users the correct way to do things (then they can choose to be lazy if they wish).

src/Examples/UnitModels/cstr_testing.ipynb
bpaul4: `</div>` is a tag to end the section, the lines above it create a bulleted list (it renders correctly if you view the file). I'll change it to match the bulleted list on line 62, since that syntax looks a bit cleaner.

src/Examples/UnitModels/cstr_testing.ipynb
bpaul4: This syntax does work, I will update all of the specifications to explicitly include units and do this going forward as well.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1107 @@
...
+   "execution_count": 6,
+   "metadata": {
+    "scrolled": false,
+    "tags": [
+     "solution"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "#Todo: Add reactor with the specifications above\n",
```
jghouse88, line 219: Given this is an example, we do not need the exercise cells here. Just plain old series of cells with testing cells will be sufficient. Tutorials have the `TODO` exercise cells.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1107 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.cooling_cost = Expression(expr=0.212e-7 * (-m.fs.R101.heat_duty[0]))  # the reaction is exothermic, so R101 duty is negative\n",
+    "m.fs.heating_cost = Expression(expr=2.2e-7 * m.fs.H101.heat_duty[0])  # the stream must be heated to T_rxn, so H101 duty is positive\n",
+    "m.fs.operating_cost = Expression(expr=(3600 * 8000 *(m.fs.heating_cost + m.fs.cooling_cost)))"
```
jghouse88, line 307: Are you assuming some downtime in a year? Why 8000 instead of 8760? Totally acceptable if you say that there is some downtime in a year.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1107 @@
...
+    "m.fs.M101.catalyst_feed.pressure.fix(1e5*pyunits.Pa)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Fixing unit model specifications\n",
+    "\n",
+    "Now that we have fixed our inlet feed conditions, we will now be fixing the operating conditions for the unit models in the flowsheet. Let us initialize H101 with an outlet temperature of 328.15 K. "
```
jghouse88, line 387: I would simply say "fix" temperature to. Initialize makes it sound as a temporary fix for initializing only. I understand you will be unfixing this later for optimization but fixing might be the correct term here.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1107 @@
...
+    "Typically in reactor sizing, the liquid volume (reaction volume) does not exactly reach the top of the reactor vessel due to overhead space, vapor draws, or pressure-release valves. We will assume that 80% of our total reactor volume (vessel size, normally used for equipment costing) is our liquid, which is a common feature of stirred-tank or batch reactors. For the CSTR, we have to define the conversion in terms of ethylene oxide as well as the CSTR reaction volume. This requires us to create new variables and constraints relating reactor properties to stream properties.  Note that the CSTR reaction volume variable (m.fs.R101.volume) does not need to be defined here since it is internally defined by the CSTR model. Additionally, the heat duty is not fixed, since the heat of reaction depends on the reactor conversion (through the extent of reaction and heat of reaction). We'll estimate 80% conversion for our initial flowsheet:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m.fs.R101.total_volume = Var(initialize=6.9225, bounds=(0, 15), units=pyunits.m**3)  # m^3\n",
```
jghouse88, line 416: Minor comment: This new var (`total_volume`) could have been avoided as we did not include the capex anywhere in the optimization. Even if we included the capex, we could have used the reactor volume var and multiplied it by a factor of 1.2 to give the 20% design margin.

jghouse88 (changes requested):
Looks great, Brandon. Minor comments. Very well done!

src/Examples/UnitModels/cstr_testing.ipynb
bpaul4: I am assuming some downtime, and 8000 is a rough estimate (about a month of downtime (~10%) per year, or 333 operating days/year). I'll add a note in the text cell where I mention this number.

src/Examples/UnitModels/cstr_testing.ipynb
```diff
@@ -0,0 +1,1107 @@
...
+      "       Volume :      12.581 : False : (None, None)\n",
+      "\n",
+      "------------------------------------------------------------------------------------\n",
+      "    Stream Table\n",
+      "                                                 Inlet     Outlet  \n",
+      "    Molar Flowrate ('Liq', 'ethylene_oxide')      58.000     5.8000\n",
+      "    Molar Flowrate ('Liq', 'water')               239.60     187.40\n",
+      "    Molar Flowrate ('Liq', 'sulfuric_acid')      0.33401    0.33401\n",
+      "    Molar Flowrate ('Liq', 'ethylene_glycol') 2.0000e-05     52.200\n",
+      "    Temperature                                   328.15     329.27\n",
```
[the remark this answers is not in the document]
bpaul4: Yes, it is fairly small (dHrxn = - 11.52 kcal/mol = -48 kJ/mol). The activation energy is small as well (-9.79 kcal/mol = -41 kJ/mol) as a result of the sulfuric acid catalyst. The numbers are taken from published literature with experimental data on EO hydrolysis to EG with this catalyst.

src/Examples/UnitModels/cstr_testing.ipynb
bpaul4: Would you suggest removing this variable? I believed it is an important point to mention regarding reactor design, and I agree a design margin may be less confusing to users than an additional variable.

jghouse88: I would @bpaul4. A comment will be sufficient or you can print the total volume to be 1.2 times the reactor volume you get.

src/Examples/UnitModels/cstr_testing.ipynb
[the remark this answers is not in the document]
jghouse88: Okies, just making sure.

jghouse88 (approved):
LGTM. Great job @bpaul4

andrewlee94 (approved):
Looks good to me.

PR #84, opened 2021-12-22
Adding PFR and StoichiometricReactor Notebook Examples

bpaul4:
## Proposed changes:
- New notebook examples for PFR and StoichiometricReactor unit models and associated index updates

### Legal Acknowledgement


notebook_index.yml
```diff
@@ -119,6 +119,8 @@ contents:
             description: Detailed examples on importing and implementing IDAES reactor models in flowsheets
             notebooks:
               - cstr: CSTR unit model with ideal property package, pure liquid phase and rate-kinetics reaction
+              - plug_flow_reactor: Plug flow reactor unit model with ideal property package, pure liquid phase and rate-kinetics reaction
```
andrewlee94, line 122: Not sure what is meant by "pure" liquid phase here. Do you mean "single"?

Also line below.

src/notebook_index.ipynb
```diff
@@ -116,8 +116,10 @@
     "<a id='examples.unitmodels.reactors'></a>\n",
     "\n",
     "#### Reactors\n",
-    "Detailed examples on importing and implementing IDAES reactor models in flowsheets\n",
+    "Detailed examples on importing and implementing IDAES reactor models\n",
     "  * [cstr](Examples/UnitModels/Reactors/cstr.ipynb) - CSTR unit model with ideal property package, pure liquid phase and rate-kinetics reaction\n",
+    "  * [plug_flow_reactor](Examples/UnitModels/Reactors/plug_flow_reactor.ipynb) - Plug flow reactor unit model with ideal property package, pure liquid phase and rate-kinetics reaction\n",
```
andrewlee94, line 121: Same comment r.e. "pure" liquid phase.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    "# Flowsheet PFR Simulation and Optimization of Ethylene Glycol Production\n",
```
andrewlee94, line 8: Whilst most readers will probably understand, you should write Plug Flow Reactor out in full the first time you use it.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+    "    - Setting variable bounds\n",
+    "    - Adding additional constraints \n",
+    "\n",
+    "\n",
+    "## Problem Statement\n",
+    "\n",
+    "Following the previous example implementing a CSTR unit model, we can alter the flowsheet to use a plug flow reactor (PFR). As before, this example is adapted from Fogler, H.S., Elements of Chemical Reaction Engineering 5th ed., 2016, Prentice Hall,  p. 157-160 with the following chemical reaction, property packages and flowsheet. Unlike a CSTR which assumes well-mixed liquid behavior, the concentration profiles will vary spacially in one dimension. In actuality, following start-up flow reactor exhibit dynamic behavior as they approach a steady-state equilibrium; we will assume our system has already achieved steady-state behavior. The state variables chosen for the property package are **molar flows of each component by phase in each stream, temperature of each stream and pressure of each stream**. The components considered are: **ethylene oxide, water, sulfuric acid and ethylene glycol** and the process occurs in liquid phase only. Therefore, every stream has 4 flow variables, 1 temperature and 1 pressure variable.\n",
```
andrewlee94, line 25: Similarly, we probably should expand CSTR the first time. It would also be nice to embed a link to the CSTR example, but I am not sure how easy that is to do with the way our docs get built (@dangunter - any ideas?).

Typo: spatially.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+    "from idaes.core.util.initialization import propagate_state\n",
+    "\n",
+    "import idaes.logger as idaeslog"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Importing required thermo and reaction package\n",
```
andrewlee94, line 105: Should write thermophysical property out in full, package should be plural.

Same for all instances of thermo below - this is a formal document so we should avoid shorthands.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+    "$C_{EO}$ - bulk concentration of ethylene oxide in moles per $m^3$ (the limiting reagent, since we assume excess catalyst and water)  \n",
+    "$k_0$ - pre-exponential Arrhenius factor per second  \n",
+    "$E_a$ - reaction activation energy in kJ per mole of ethylene oxide consumed  \n",
+    "$R$ - gas constant in J/mol-K  \n",
+    "$T$ - reactor temperature in K\n",
+    "\n",
+    "These calculations are contained within the property, reaction and unit model packages, and do not need to be entered into the flowsheet. More information on property estimation may be found below:\n",
+    "\n",
+    "ParamEst parameter estimation: https://idaes-pse.readthedocs.io/en/stable/user_guide/workflow/data_rec_parmest.html?highlight=paramest  \n",
+    "HELMET thermodynamic estimation: https://idaes-pse.readthedocs.io/en/stable/user_guide/modeling_extensions/surrogate/helmet/index.html  \n",
```
andrewlee94, line 125: I am not sure about referencing HELMET and RIPE right now - the code is an absolute mess, and is not tested so we don't even know if they actually work. Both of those are likely to be rebuilt next year, or pushed of to the side somewhere,

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+   "source": [
+    "m = ConcreteModel()\n",
+    "m.fs = FlowsheetBlock(default={\"dynamic\": False})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We now need to add the property packages to the flowsheet. Unlike Module 1, where we only had a thermo property package, for this flowsheet we will also need to add a reaction property package. We will use the Generic Property and Generic Reaction Frameworks; more information may be found on these methods at https://idaes-pse.readthedocs.io/en/1.8.0/user_guide/components/property_package/index.html."
```
andrewlee94, line 166: Again, it would be nice to link to Module 1 here if possible.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+   "source": [
+    "m.fs.R101 = PFR(\n",
+    "            default={\"property_package\": m.fs.thermo_params,\n",
+    "                     \"reaction_package\": m.fs.reaction_params,\n",
+    "                     \"has_heat_of_reaction\": True,\n",
+    "                     \"has_heat_transfer\": True,\n",
+    "                     \"has_pressure_change\": False,\n",
+    "                     \"transformation_method\": \"dae.finite_difference\",\n",
+    "                     \"transformation_scheme\": \"BACKWARD\",\n",
+    "                     \"finite_elements\": 1,\n",
```
andrewlee94, line 224: This is not a good idea - a PFT with one element is just a CSTR. It doesn't cost us much to use more elements, so I would suggest we use a few (5 to 10 maybe).

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+    "                     \"length_domain_set\": [0, 1]})"
```
andrewlee94, line 225: I don't think we should include the `length_domain_set` in this example - it is an advanced option and you need to be careful when using it.

Make sure you update any text above if you change these.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+    "m.fs.eg_prod = Expression(expr=pyunits.convert(m.fs.R101.outlet.flow_mol_phase_comp[0, \"Liq\", \"ethylene_glycol\"]\n",
+    "                                               *m.fs.thermo_params.ethylene_glycol.mw, # MW defined in properties as kg/mol\n",
+    "                                               to_units=pyunits.Mlb/pyunits.yr)) # converting kg/s to MM lb/year"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, let us add expressions to compute the reactor cooling cost (\\\\$/s) assuming a cost of 0.212E-4 \\\\$/kW, and the heating utility cost (\\\\$/s) assuming 2.2E-4 \\\\$/kW. Note that the heat duty is in units of watt (J/s). The total operating cost will be the sum of the two, expressed in \\\\$/year assuming 8000 operating hours per year (~10\\% downtime, which is fairly common for small scale chemical plants):"
```
andrewlee94, line 296: Check the formatting here. On Github's renderer, "\\\\$/kW, and the heating utility cost (\\\\$/s) assuming 2.2E-4 " was all in math model (but that might just be Github).

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1147 @@
...
+      "2021-12-22 13:37:46 [INFO] idaes.init.fs.H101.control_volume.properties_in: Property initialization: optimal - Optimal Solution Found.\n",
+      "2021-12-22 13:37:46 [INFO] idaes.init.fs.H101.control_volume.properties_out: Starting initialization\n",
+      "2021-12-22 13:37:46 [INFO] idaes.init.fs.H101.control_volume.properties_out: Property initialization: optimal - Optimal Solution Found.\n",
+      "2021-12-22 13:37:46 [INFO] idaes.init.fs.H101.control_volume: Initialization Complete\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.H101: Initialization Complete: optimal - Optimal Solution Found\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.R101.control_volume.properties: Starting initialization\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.R101.control_volume.properties: Property initialization: optimal - Optimal Solution Found.\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.R101.control_volume.reactions: Initialization Complete.\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.R101.control_volume: Initialization Complete\n",
+      "2021-12-22 13:37:47 [INFO] idaes.init.fs.R101: Initialization Complete: optimal - Optimal Solution Found\n"
```
andrewlee94, line 499: We should add an `InitializationError` (a new exception type) to the property package so that if something breaks in the future we will see exceptions hereto let us know. `InitializationErrors` are currently being added to the unit models in a draft PR in idaes-pse.

andrewlee94 (changes requested):
I haven't look at the stoichiometric reactor notebook yet, but I am guessing many of the comments for the PFR notebook will apply there too.

notebook_index.yml
bpaul4: The three rate-reactor examples exhibit liquid phase behavior only (no vapor).

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: The code renders correctly in the notebook, as well as the chemical reaction equation.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: I will take a look at the PR.

bpaul4:
Thank you @andrewlee94 for your comments, I will review and update all three notebooks for any changes.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: Looking at the `notebook_index.ipynb` file, there is a link to the local host file - I will attempt embedding it behind the text, and do the same for Module 1 below if it is successful.

bpaul4: Update: I successfully added an internal link by copying the "localhost" link generated in `notebook_index.ipynb` and editing the example text to [text to display as blue hyperlink](URL the link opens when clicked)

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: Would you recommend that I remove these references from the examples?

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: See earlier comment on embedding links in Markdown.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
bpaul4: The PFR generates a degree of freedom for each finite element, and the initialization returns an infeasible solution when I fix the heat duty of each element to zero and unfix all other reactor specifications. I'm referring to the [test_pfr](https://github.com/IDAES/idaes-pse/blob/main/idaes/generic_models/unit_models/plug_flow_reactor.py) example as I troubleshoot how to satisfy the degrees of freedom.

andrewlee94: That indicates that there is a fundamental issue with the formulation of the problem then, which needs to be addressed before we merge this.

bpaul4:
@andrewlee94 I've made the suggested changes in all three rate-reactor notebook examples, including adding the two links. Additionally, I reformulated the initial plug-flow reactor problem which solves correctly now.

src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb
```diff
@@ -0,0 +1,1160 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "m.fs.R101 = PFR(\n",
+    "            default={\"property_package\": m.fs.thermo_params,\n",
```
andrewlee94, line 217: I will show how to set the number of elements here (even if you use the default value) - it is important for users to know how to do this. We should also show how to set the discretization method (both arguments).

bpaul4: I will add these arguments back in.

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
```diff
@@ -0,0 +1,1073 @@
...
+    "from idaes.core import FlowsheetBlock\n",
+    "from idaes.generic_models.properties.core.generic.generic_property import (\n",
+    "        GenericParameterBlock)\n",
+    "from idaes.generic_models.properties.core.generic.generic_reaction import (\n",
+    "        GenericReactionParameterBlock)\n",
+    "from idaes.generic_models.unit_models import (Mixer,\n",
+    "                                              Heater,\n",
+    "                                              StoichiometricReactor)\n",
+    "\n",
+    "from idaes.core.util.constants import Constants\n",
```
andrewlee94, line 94: I do not think this is used either.

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
```diff
@@ -0,0 +1,1073 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Importing required thermophysical and reaction packages\n",
+    "\n",
+    "The final set of imports are to import the thermophysical and reaction packages. We have created a custom thermophysical package that support ideal vapor and liquid behavior for this system, and in this case we will restrict it to ideal liquid behavior only.\n",
+    "\n",
+    "The reaction package here assumes Arrhenius kinetic behavior for the StoichiometricReactor, for which $k_0$ and $E_a$ are known *a priori* (if unknown, they may be obtained using one of the parameter estimation tools within IDAES).\n",
```
andrewlee94, line 110: For the stoichiometric reactor, we do not use the rate constant at all so I wonder if it would be better to leave most of this out. All that gets used from the reaction package is the definition of the reactions and their stoichiometry.

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
```diff
@@ -0,0 +1,1073 @@
...
+   "outputs": [],
+   "source": [
+    "m.fs.H101.outlet.temperature.fix(328.15*pyunits.K)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will need to specify both initial conversion and heat duty values (these are the only two free variables to choose from). Since heat duty and the outlet reactor temperature are interdependent, we can choose to specify  this quantity instead. While the reaction kinetic parameters exist in the property package, we also do not need to add a rate constant expression since generation is explicitly defined through the conversion/yield. Note that our initial problem will solve with zero *temperature change* but will be infeasible with zero *heat duty*; this is due to the heat of reaction enforced by allowing heat transfer and mandating a non-zero conversion."
```
andrewlee94, line 398: I wonder if it is worth mentioning that there are aa number of ways the reaction extent could be defined; e.g. conversion or yield.

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
```diff
@@ -0,0 +1,1073 @@
...
+    "assert value(m.fs.R101.outlet.temperature[0])/1E2 == pytest.approx(3.2815, abs=1e-3)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Optimizing Ethylene Glycol Production\n",
+    "\n",
+    "Now that the flowsheet has been squared and solved, we can run a small optimization problem to minimize our production costs. Suppose we require at least 200 million pounds/year of ethylene glycol produced and 90% conversion of ethylene oxide, allowing for variable reactor volume (considering operating/non-capital costs only) and reactor temperature (heater outlet)."
```
andrewlee94, line 721: The mention of reactor volume here seems out of place, as a stoichiometric reactor by its nature does not depend on volume. I am guessing this is left over from the previous examples.

andrewlee94 (commented):
I did a quick read through of the stoichiometric reactor example this time and only had a few comments. Most of them are imports I don't think you need, plus a couple of things that don't make much sense for a stoichiometric reactor, but were left over from the CSTR/PFR example.

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
bpaul4: Possibly, conversion is typically more convenient since it is independent of the reaction stoichiometry. There are multiple ways to specify yield (mol EG/mol EO, kg EG/kg EO, or volumetric units) while conversion is more straightforward always being on 0-100% for a given basis. I should also emphasize that conversion is chosen with respect to a specific reactant.

bpaul4:
@andrewlee94 I cleaned up the notebooks and updated the PR

src/Examples/UnitModels/Reactors/stoichiometric_reactor_testing.ipynb
andrewlee94: Not quite - conversion also depends on the stoichiometric basis chosen, as it can be specified with respect to any of the reactants as well, and as reactants are not always fed stoichiometrically a 100% conversion of one reactant is not always equivalent to a 100% conversion of the other (although conventionally we  would use the limiting reagent as the basis, it is just as valid to do it another way).

However, my main point here is that we should highlight that IDAES supports any of these approaches and bases for specifying the reaction extent (including directly setting the extent), and that we are going to use conversion of species X for this example.

bpaul4: The section has been updated with the details discussed above.

bpaul4:
@MAZamarripa @jghouse88 This PR is awaiting an additional review. Please let me know if you will have time to review this, or if there is another reviewer I may add. Thanks!

radhakrishnatg (approved):
Looks good to me.

andrewlee94 (approved):
I only did a quick skim, so I will trust that my prior reviews were sufficient.

PR #98, opened 2022-02-15
Adding Methanol Synthesis Flowsheet Example Notebook

bpaul4:
## Proposed changes:
- Adding methanol synthesis example notebook to EXAMPLES-PSE repository

### Legal Acknowledgement


lbianchi-lbl:
@bpaul4 If you'd like to test this PR with IDAES/idaes-pse#711 before that one is merged into `idaes-pse:main`, you should be able to just set the proper value here: https://github.com/IDAES/examples-pse/blob/79dfed7ade2a352bf9d847bbd155ed40bf6d4559/.github/workflows/main.yml#L17, i.e.:

```yaml
env:
  IDAES_CONDA_ENV_NAME_DEV: examples-pse-dev
  IDAES_PSE_REQUIREMENT: idaes-pse[prerelease] @ git+https://github.com/IDAES/idaes-pse@refs/pull/711/head
``

File: src/Examples/Flowsheets/methanol_synthesis.ipynb [hunk 1]
```diff
@@ -0,0 +1,1749 @@
...
+    "print()\n",
+    "print('Solving optimization problem...')\n",
+    "opt_res = solver.solve(m, tee=True)\n",
+    "assert opt_res.solver.termination_condition == TerminationCondition.optimal\n",
+    "print('Optimal solution process results:')\n",
+    "report(m)"
+   ]
+  },
+  {
+   "cell_type": "code",
```
andrewlee94, line 1705: It is hard to tell via the GitHub interface, but this appears to be missing the tags to indicate that this is a testing cell and should not be included in the public facing version,

andrewlee94 (changes requested):
As a general comment, I think there needs to be more guidance to the user if this is to be a proper example. Each step should explain what is happening, why and what the expected outcome is.

ksbeattie:
@andrewlee94, @MAZamarripa this is ready for re-review.

andrewlee94 (approved):
I took a quick look and only had a couple of minor observations:

1. A summary upfront to explain the purpose of the example (in a sentence or two) would probably be valuable. This would help inform users of whether this example is of interest to them.

2. Do we really need 500 iterations for IPOPT? Unless there is a good reason, I think we should be using a more reasonable value for this. Otherwise, readers will start to think 500 iterations is expected/acceptable, when really it should solve in less than 100 in most cases.

bpaul4:
@MAZamarripa this is ready for your review when you have a moment. Thank you!

MAZamarripa (approved):
lgtm

PR #100, opened 2022-02-17
Update scaling to fix failing example notebooks

bpaul4:
## Fixes # .
- Added scaling to `src/Examples/Advanced/CustomProperties/air_separation_case_study.ipynb`
- Updated scaling for `src/Examples/Advanced/CustomProperties/Hydrocarbon_Processing_example.ipynb`

## Proposed changes:
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #96, opened 2022-02-14
Several examples tests failing


eslickj:
There are some examples test failing.  I'm having a hard time sorting out what's failing and how, but I suspect it may have something to do with the doc reorg or updating the Pyomo version.  Not sure who to assign this too, so we can try to update that later.

From doc build:

```
docs_test/Examples/Tools/degeneracy_hunter.rst:234: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:234: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:409: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:409: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:409: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:712: ERROR: Undefined substitution referenced: "residual".
docs_test/Examples/Tools/degeneracy_hunter.rst:712: ERROR: Undefined substitution referenced: "residual".
```

```
docs_test/Examples/Tools/data_management_framework.rst:514: ERROR: Unknown target name: "id".
docs_test/Examples/Tools/data_management_framework.rst:576: ERROR: Unknown target name: "id".
```

```
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
docs_test/Examples/SurrMod/PySMO/PySMO_basics.rst:378: ERROR: Unknown target name: "x".
```

```
TypeError: object NoneType can't be used in 'await' expression
[...]

PR #101, opened 2022-02-18
Temporarily move notebooks causing CI failures to separate directory

lbianchi-lbl:
## Motivation

- Workaround for #96 as suggested in #100

## Proposed changes:
- Move notebooks with known issue causing CI failures to `_skip` directory so that they are not run in CI
  - The solution of moving the notebooks to a separate directory is because the current build configuration is directory-based, and we'd like to continue testing the other notebooks that happen to be in the same directory as the ones being skipped

### Legal Acknowledgement


andrewlee94 (approved):
Looks good to me (at least as a temporary fix).

bpaul4:
@andrewlee94, if you are fine with leaving the rest of the doctree alone and just moving the broken examples for now, I will approve this so we can start merging queued PRs.

lbianchi-lbl:
@bpaul4 that's a good point, thanks for spotting it. I've just updated the notebook index; it seems to be already commented out from the `index.rst` file but we can deal with that later if that's not the case.

andrewlee94 (approved):

PR #105, opened 2022-03-03
Remove hardcoded value for idaes-pse requirement from requirements file

lbianchi-lbl:
## Motivation
- Having the `idaes-pse` requirement inside the `requirements.txt` file makes it harder to change it programmatically when needed (e.g. to use a released version of `idaes-pse` instead of the default)
- If `requirements.txt` contains all other requirements except the one for `idaes-pse`, it's easy to add it before installation with e.g. `echo "idaes-pse[prerelease] @ git+ssh://my-git-server:idaes-pse.git" >> requirements.txt`

## Proposed changes:
- Remove the requirement for `idaes-pse` from `requirements.txt`
- Tweak the GHA workflow and `setup-idaes` actions to configure and append the `idaes-pse` requirement before installation

### Legal Acknowledgement

### DIFF
--- a/.github/actions/setup-idaes/action.yml
+++ b/.github/actions/setup-idaes/action.yml
@@ -24,6 +24,9 @@ runs:
     - name: Install idaes and dependencies
       shell: bash -l {0}
       run: |
+        echo '::group::Contents of requirements.txt'
+        cat requirements.txt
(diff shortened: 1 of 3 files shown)


ksbeattie (approved):
LGTM

PR #110, opened 2022-04-11
Add scaling to reduce example runtime

bpaul4:
## Fixes # .
In recent workflow runs on GitHub Actions, two example notebooks have presented recurring failures due to long runtimes or solver failures. These issues are most likely due to bad initialization and poor scaling, and this update remedies those concerns.

## Proposed changes:
- Added scaling to `src/Examples/Advanced/CustomProperties/air_separation_case_study.ipynb`
- Added scaling to `src/Examples/Advanced/CustomProperties/Hydrocarbon_Processing_example.ipynb`
- Re-generated `src/Examples/Advanced/CustomProperties/HI_init.json.gz` for new scaled problem
- Associated changes to doctree files and removing notebooks from `src/_skip` folder

### Legal Acknowledgement


bpaul4:
It appears that with the new scaling the air separation notebook initializes with no errors, and the HC processing example runs in about half the time (700 seconds compared to 1300 seconds with no scaling). @andrewlee94 I would appreciate your thoughts on improved scaling factors to lower the runtime below the 600 second threshold.

andrewlee94:
@bpaul4 First, it seems the hydrocarbon example is failing on Windows, so the issue there has not been resolved yet (and in fact this is the main issue with these models in general - they will work on one OS but final on the other).

Second, on the topic of time, is the 700 seconds the total time to run the notebook. or just the time spent in IPOPT?

bpaul4:
To address your first point, the hydrocarbon example runs with no issues on Windows - it just takes too long and the CI client times out. The tests pass on my local machine, and once again I'm wondering if it would be best to update `build-ci.py` and bump the timeout from 600 to 1200 seconds to match `build.py`. For your second point, 700 seconds is the total time to run the notebook, per the notebook output the solver spends 30 IPOPT seconds on the post-initialization solve and 150 IPOPT seconds on the optimization solve.

andrewlee94:
@bpaul4 OK - that is a very different problem then and one that scaling  is not necessarily the fix for (although 150 seconds in IPOPT is a bit). The bigger issue however is that it takes ~400 seconds to build the model which is already 2/3rds of the allotted time for the test.

We could allow more time for the tests, but we should be careful as we do want to keep our test times from growing out of hand. One option would be to add a specific allowance for this test alone (if that is possible), or maybe make this a test that is only run on a scheduled basis (rather than on all PRs), which would let us relax the time requirement a little (but not too much).

bpaul4:
If we are going to add a flag, switch, etc to allow individual tests to overwrite the default timeout value, I believe we would need to modify this section (https://github.com/IDAES/examples-pse/blob/main/build.py#L455) in order to leave `build-ci.py` generic for all tests. One idea I have is to create a new CELL_TAG "run_long" in specific notebooks, and modify the `build.py` script to default to `timeout=1200 s` if this tag exists. Anything running longer than 1200 s should fail so we know about it, and 1200 s is a good choice since that's what we use for the local build script.

lbianchi-lbl:
> If we are going to add a flag, switch, etc to allow individual tests to overwrite the default timeout value, I believe we would need to modify this section (https://github.com/IDAES/examples-pse/blob/main/build.py#L455) in order to leave `build-ci.py` generic for all tests. One idea I have is to create a new CELL_TAG "run_long" in specific notebooks, and modify the `build.py` script to default to `timeout=1200 s` if this tag exists. Anything running longer than 1200 s should fail so we know about it, and 1200 s is a good choice since that's what we use for the local build script.

@bpaul4 Thanks for your work on this. Regarding this proposed solution, I think this is reasonable in general, however we'd still need to address the need of excluding the long-running notebooks from the build altogether for cases when we want to keep the overall runtime short (i.e. CI jobs that run on every PR, either in this repo or in `idaes-pse`). We're trying to figure out what would be the most effective way to achieve this, considering the limited resources that make it difficult to develop a more "proper" solution.

bpaul4:
@lbianchi-lbl @andrewlee94 thank you for providing your expertise in this area. I am seeking to expand my YAML knowledge and am unfamiliar with some of the syntax at the moment, but it could be efficient to add a `schedule:` trigger to run long notebooks periodically rather than on every PR. For example, run all notebooks once a week with a timeout of 1200 s (or longer), and on normal PR and nightly runs exclude the long notebooks and keep the timeout at 600 s.

andrewlee94:
@bpaul4 I think you might be confusing a few things here - YAML is just a format for storing information (basically a dict). In this case, we are using the YAML file to store the information for setting up the CI run, which is a whole different infrastructure, and that is where the issue lies.

However, this is all part of a larger question about what tests get run when and how, and the current infrastructure is rather cumbersome and limiting in that sense.

bpaul4:
@andrewlee94 @lbianchi-lbl I've moved the notebooks back to the `_skip` directory while we debate integrating these examples into the CI workflows.

andrewlee94 (approved):
I think this looks good - there are a few things that should be fixed at some point but I think we can deal with those later.

bpaul4:
@lbianchi-lbl @agarciadiego when you have a moment, could you review this? Thanks!

agarciadiego (approved):
LGTM

PR #111, opened 2022-04-13
Adding Skeleton Unit Model Example

bpaul4:
## Fixes # .
- New notebook example for Skeleton Unit Model and associated index updates

## Proposed changes:
-

### Legal Acknowledgement


src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+   "metadata": {},
+   "source": [
+    "# IDAES Skeleton Unit Model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook demonstrates usage of the IDAES Skeleton Unit Model, which provides a generic \"bares bones\" unit for user-defined models and custom variable and constraint sets. To allow maximum versatility, this unit may be defined as a surrogate model or a custom equation-oriented model. Users must add ports and variables that match connected models, and this is facilitated through a provided method to add port-variable sets.\n",
```
andrewlee94, line 14: Type: bares -> bare

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "\n",
+    "For users who wish to train surrogates with IDAES tools and insert obtained models into a flowsheet, see the more versatile `SurrogateBlock()` for streamlined integration."
```
andrewlee94, line 16: This should be reworded. Just direct users to the general surrogate modeling docs rather than `SurrogateBlock`. For one, `SurrogateBlocks` do not handle the training of the surrogate, so the above statement is misleading at face value.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "from idaes.models.unit_models import SkeletonUnitModel\n",
+    "from idaes.core.util.model_statistics import degrees_of_freedom\n",
+    "from pyomo.util.check_units import assert_units_consistent"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For demonstrative purposes, we will build a simple model manually defining state variables relations for two generic components. In this model, the variables specify an FcTP system where molar flow of each component, temperature and pressure are selected as state variables. Users can use this structure to write custom relations between inlet and outlet streams; for example, defining outlet pressure using an explicit correlation or outlet flow of a specific component from a known accumulation or loss in the system."
```
andrewlee94, line 51: Put `FcTP` within backticks (or something similar) to give it a different formatting to the main text to highlight it.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pytest\n",
+    "from pyomo.environ import (check_optimal_termination,\n",
```
andrewlee94, line 33: I wonder if we should transition to the Pyomo standard of just doing `import pyomo.environ as pyo` here.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "For demonstrative purposes, we will build a simple model manually defining state variables relations for two generic components. In this model, the variables specify an FcTP system where molar flow of each component, temperature and pressure are selected as state variables. Users can use this structure to write custom relations between inlet and outlet streams; for example, defining outlet pressure using an explicit correlation or outlet flow of a specific component from a known accumulation or loss in the system."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = ConcreteModel()\n",
```
andrewlee94, line 60: This needs to be broken up, with descriptions of each part explaining what is being done and why and referring back to the introduction to the model.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "    return m.flow_comp_out[t, i] == m.flow_comp_in[t, i]\n",
+    "m.fs.skeleton.eq_flow = Constraint(m.fs.time, m.fs.skeleton.comp_list, rule=rule_flow)\n",
+    "\n",
+    "# note that all terms need to have consistent units - the \"10\" terms could be defined as separate variables with units\n",
+    "m.fs.skeleton.eq_temperature = Constraint(expr=m.fs.skeleton.temperature_out[0]\n",
+    "                                          == m.fs.skeleton.temperature_in[0] + 10*pyunits.K)\n",
+    "m.fs.skeleton.eq_pressure = Constraint(expr=m.fs.skeleton.pressure_out[0]\n",
+    "                                       == m.fs.skeleton.pressure_in[0] - 10*pyunits.kPa)\n",
+    "\n",
+    "# dictionaries relating state properties to custom variables\n",
```
andrewlee94, line 90: This part especially needs more description - what are these `dicts`, how they will be used and what they must contain.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+   ],
+   "source": [
+    "print(degrees_of_freedom(m))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By specifying four input values, we obtain a square problem and can initialize and solve the flowsheet similar to other IDAES unit models. While importing IDAES unit models requires fully consistent physical unit sets, the Skeleton model allows users to only include a few specific equations and easily allows custom physical units as well. As shown below, the model only requires unit consistency within equations as no other unit, block or property modules are called:"
```
andrewlee94, line 130: 1. This paragraph seems to cover two separate topics, and I think it should be broken into two with more discussion. The first part deals with degrees of freedom, and the second part with consistency of "units" (see my third point).

2.I think you can say more about degrees of freedom, such as that we would expect 4 degrees of freedom (referring back to the introduction of the model as appropriate). From that, explicitly call out the 4 things we intend to fix to make the problem square.

3. I am not sure what the last part of this is trying to say - how does "importing IDAES unit models requires fully consistent physical unit sets", and how does the `SkeletonModel` address this?

Notably, I think you are using the term "units" in multiple senses in this one paragraph. The first part is saying the "units don't need to be consistent", and then the final sentence says "he model only requires unit consistency within equations" which would appear to be contradictory. What I think you mean is that `SkeletonModels` do not require consistent thermophysical properties, but consistency of units of measure must still be maintained. This needs to be clarified, and you should choose your words carefully.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+      "Number of Lagrangian Hessian evaluations             = 0\n",
+      "Total CPU secs in IPOPT (w/o function evaluations)   =      0.000\n",
+      "Total CPU secs in NLP function evaluations           =      0.000\n",
+      "\n",
+      "EXIT: Optimal Solution Found.\n"
+     ]
+    }
+   ],
+   "source": [
+    "m.fs.skeleton.inlet.flow_mol_comp[0, \"c1\"].fix(2*pyunits.mol/pyunits.s)\n",
```
andrewlee94, line 207: This should also be broken up with some description of each step.

Of especial note, you need to discuss what happens when `m.fs.skeleton.initialize()` is called. Given that all of the content of the `SkeletonModel` is user written, the pre-written `initialize` method is going to be very simplistic and users need to be aware of exactly what its limitations are and we should probably provide some pointers on what to do if this fails (no details, but point to other docs (which might not exist yet).

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "m.fs.skeleton.inlet.flow_mol_comp[0, \"c2\"].fix(2*pyunits.mol/pyunits.s)\n",
+    "m.fs.skeleton.inlet.temperature.fix(325*pyunits.K)\n",
+    "m.fs.skeleton.inlet.pressure.fix(200*pyunits.kPa)\n",
+    "\n",
+    "assert degrees_of_freedom(m) == 0\n",
+    "\n",
+    "m.fs.skeleton.initialize()\n",
+    "\n",
+    "solver = SolverFactory('ipopt')\n",
```
andrewlee94, line 216: You should use the IDAES `get_solver` method here.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "results = solver.solve(m, tee = True)\n",
+    "\n",
+    "assert check_optimal_termination(results)\n",
+    "assert_units_consistent(m)\n",
```
andrewlee94, line 220: I think most of these asserts would be better as a test cell, and that we should instead use the `display()` method for output to the user.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "assert value(m.fs.skeleton.outlet.flow_mol_comp[0, \"c2\"]) == pytest.approx(2, abs=1e-3)\n",
+    "assert value(m.fs.skeleton.outlet.temperature[0]) == pytest.approx(335, abs=1e-3)\n",
+    "assert value(m.fs.skeleton.outlet.pressure[0]) == pytest.approx(190, abs=1e-3)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Skeleton model allows for a custom, user-defined initialization scheme as well, as demonstrated below. Users can similarly introduce custom variable and constraint scaling using IDAES scaling tools (not demonstrated in this example):"
```
andrewlee94, line 232: This needs far more discussion and explanation. What are you doing here, what is different to the first example and what does it demonstrate to the user?

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.8.12"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
```
andrewlee94, line 420: I think there needs to be some sort of summary or conclusion at the end to remind users of what they just read.

andrewlee94 (changes requested):
I think there needs to be a lot more discussion and explanation to tell users not only how to use a `SkeletonModel`, but also when and why they would want to.

I am also not sure if you have done all the necessary steps to get this example integrated with the build and test harnesses (I seem to recall there is another file you need to add the notebook to).

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,420 @@
...
+    "# m.fs.time is a pre-initialized Set belonging to the FlowsheetBlock; for dynamic=False, time=[0]\n",
+    "m.fs.skeleton.flow_comp_in = Var(m.fs.time, m.fs.skeleton.comp_list, initialize=1.0, units=pyunits.mol/pyunits.s)\n",
+    "m.fs.skeleton.temperature_in = Var(m.fs.time, initialize=298.15, units=pyunits.K)\n",
+    "m.fs.skeleton.pressure_in = Var(m.fs.time, initialize=101, units=pyunits.kPa)\n",
+    "\n",
+    "# output vars for skeleton\n",
+    "m.fs.skeleton.flow_comp_out = Var(m.fs.time, m.fs.skeleton.comp_list, initialize=1.0, units=pyunits.mol/pyunits.s)\n",
+    "m.fs.skeleton.temperature_out = Var(m.fs.time, initialize=298.15, units=pyunits.K)\n",
+    "m.fs.skeleton.pressure_out = Var(m.fs.time, initialize=101, units=pyunits.kPa)\n",
+    "\n",
```
radhakrishnatg, line 76: From my experience, one of the limitations of declaring the flow, temperature and pressure variables this way is the following. In most cases, the skeleton unit model will be used in the context of a larger flowsheet. If the FcTP variables are defined in this manner, users will not be able to use `TransformationFactory` and `propagate_state`. They need to manually add the equality constraints that `TransformationFactory` adds, need to initialize the inlet state manually, and propagate the outlet state manually. So, I suggest the following. Load a thermodynamic package (any package will do), build a state block for the inlet and the outlet. 
```python
    m.fs.skeleton.properties_in = m.fs.prop_pack.build_state_block(m.fs.time, default={...})
    m.fs.skeleton.properties_out = m.fs.prop_pack.build_state_block(m.fs.time, default={...})
```
Then, add inlet and outlet ports:
```python
   # Add ports
    m.fs.skeleton.add_port(name="inlet", block=m.fs.skeleton.properties_in, doc="Inlet Port")
    m.fs.skeleton.add_port(name="outlet", block=m.fs.skeleton.properties_out, doc="Outlet Port")
```

This way, the user will be able to use `TransformationFactory` and `propogate_methods`, and the code will be clean. Also, fixing the inlet state will look familiar to the users: `m.fs.skeleton.inlet.temperature.fix(500)`

radhakrishnatg (changes requested):
Everything looks good. I just have one suggestion. Let me know your thoughts on that. 
Also, please delete the output cells. Make sure that only code and markdown cells are remaining in the notebook.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
andrewlee94: @radhakrishnatg Could you expand on why `TransformationFactory` and `propagate_state` can not be used? As long as these are the state variables used in the rest of the flowsheet I see no reason why this wouldn't work. OF course, this assumes `FcTP` are the state variables used in the rest of the flowsheet which probably should be emphasized more (hence the need for more description of the example).

For this example, I don't think there is a need to use actual `StateBlocks`, however I do think it would be worth mentioning in the conclusion (which is currently not there) that users can include `StateBlocks` or any other IDAES components within the `SkeletonModel` as desired.

radhakrishnatg: @andrewlee94 I want to rephrase what I said. They can be used, but getting them to work is a bit tricky. 
It has been a while, so I don't remember all the details. From what I remember, the keys in the inlet and outlet ports must be the same for `TransformationFactory` and `propogate_state` to work. Even though I had the same keys, I could not get the indices on the state variables to match or something like that. At least, at that time, I struggled a lot with it. I tried many things, but nothing worked. In the end, I just gave up and started using the state blocks. That made things easier for me.

andrewlee94: OK, that is a problem specific issue then. Yes, the contents of the `Port` need to match exactly, but as long as that is the case then it does not matter how you construct the Vars that are contained in it.

For the purposes of this example, I think the current approach is fine, but we should emphasize that `Ports `need to match whatever they will connect with and that it is the users responsibility to make this work.

radhakrishnatg: Okay. That works. I feel that it would be useful to include an example that shows how to make ports that are compatible with the inlet/outlet ports of the remaining units. This example would have saved a lot of time and effort for me back then.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
bpaul4: The current syntax matches what is implemented in other examples. What are your preferences regarding this and other notebooks?

bpaul4:
While I troubleshoot why the broken image link occurs in the CI directory, if you have some time @andrewlee94 @radhakrishnatg I would appreciate your feedback on the notebook itself. Thanks!

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "# flux equation (gas constant is defined as J/mol-K)\n",
+    "\n",
+    "def rule_flux(pervap, t, p, i):\n",
+    "    return pervap.permeate.flow_mol_phase_comp[t, p, i] == (pervap.permeance[t, p, i] *\n",
+    "                                                            exp(-pervap.energy_activation[t, p, i] /\n",
+    "                                                                (Constants.gas_constant *\n",
+    "                                                                 pervap.inlet.temperature[t]))\n",
+    "                                                           ) * pervap.area[t]\n",
+    "m.fs.pervap.eq_flux = Constraint(m.fs.time, m.fs.pervap.phase_list, m.fs.pervap.comp_list, rule=rule_flux)\n",
```
radhakrishnatg, line 207: Not important. Technically, this is not the equation for flux, but for the actual permeate flowrate. In order to get the flux, you need to divide the RHS by area. So, if possible, please change the description and/or the name of the equation.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "    # Callback for user provided initialization sequence\n",
+    "    unit.eq_flux.deactivate()\n",
+    "    unit.eq_cond.deactivate()\n",
+    "    unit.eq_flow.deactivate()\n",
+    "    assert degrees_of_freedom(unit) == 0\n",
+    "    if degrees_of_freedom(unit) == 0:\n",
+    "        res = solver.solve(m)\n",
+    "    unit.eq_flux.activate()\n",
+    "    unit.eq_cond.activate()\n",
+    "    unit.eq_flow.activate()\n",
```
radhakrishnatg, line 401: I do not understand what this initialization method is trying to do. You are deactivating all the pervaporation model constraints. It is not clear how the degrees of freedom become zero. Are you manually fixing the permeate and retentate flowrates somewhere? And, after you activate the equations back, how do you ensure that the solution it found earlier satisfies all the three constraints? 

I suspect the following. After you deactivate all the constraints, there is nothing to be solved, so the lines
```python
if degrees_of_freedom(unit) == 0:
    res = solver.solve(m)
```
will be solving an empty model. You can check this by printing the IPOPT log output and look at the number of variables and constraints. Let me know if you would like to discuss this in detail.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "print('Permeate water flow: ', value(m.fs.pervap.permeate.flow_mol_phase_comp[0, \"Liq\", \"water\"]), ' mol/s')\n",
+    "print('Permeate ethylene glycol flow: ', value(m.fs.pervap.permeate.flow_mol_phase_comp[0, \"Liq\", \"ethylene_glycol\"]), ' mol/s')\n",
+    "print('Permeate temperature: ', value(m.fs.pervap.permeate.temperature[0]), ' K')\n",
+    "print('Permeate pressure: ', value(m.fs.pervap.permeate.pressure[0]), ' Pa')\n",
+    "print('Retentate water flow: ', value(m.fs.pervap.retentate.flow_mol_phase_comp[0, \"Liq\", \"water\"]), ' mol/s')\n",
+    "print('Retentate ethylene glycol flow: ', value(m.fs.pervap.retentate.flow_mol_phase_comp[0, \"Liq\", \"ethylene_glycol\"]), ' mol/s')\n",
+    "print('Retentate temperature: ', value(m.fs.pervap.retentate.temperature[0]), ' K')\n",
+    "print('Retentate pressure: ', value(m.fs.pervap.retentate.pressure[0]), ' Pa')\n",
+    "\n",
+    "# separation factor for results analysis - x is inlet water fraction, y is permeate water fraction\n",
```
radhakrishnatg, line 477: Place the remaining lines from here in a new cell.

radhakrishnatg (changes requested):
It looks good overall. I just have three comments. Also, one question about the figure. Did you draw the figure or is it a snapshot taken from a textbook? If it is the latter, we cannot do that without the permission of the publisher.

bpaul4:
Thank you for your review comments @radhakrishnatg . The figure is taken from the paper, I agree with your point and will draw my own diagram to use instead of the paper diagram. I'll address your formatting suggestions as well.

Regarding the initialization, leaving the three constraints unspecified results in 3 DOF and IPOPT will not solve the unit. I tested fixing the permeate flow, retentate flow and condensation duty instead and the problem solves with no issues.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,860 @@
...
+    "assert value(m.fs.pervap.cond_duty[0]) == pytest.approx(5813, rel=1e-3)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# 3. Optimization\n",
+    "\n",
+    "Suppose we wish to characterize the membrane behavior by calculating the maximum inlet water mole fraction allowing a separation factor of at least 100 (typical value for high-efficiency separation processes such as gas separation of CO2/N2). We need to add some lower bounds on the flow variables to ensure physically-sound solutions; alternatively, we could have specified these bounds while defining the variables above. We can quickly modify and resolve the model, and check some key results:"
```
radhakrishnatg, line 614: You mentioned bounds in the description, but the code doesn't set bounds on any variables.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,860 @@
...
+      "Retentate temperature:  298.15  K\n",
+      "Retentate pressure:  101000.0  Pa\n"
+     ]
+    }
+   ],
+   "source": [
+    "# print results\n",
+    "m.fs.M101.report()\n",
+    "\n",
+    "print('Question for reviewers: Is there a way to add report() method with a stream table?')\n",
```
radhakrishnatg, line 764: See comment above and delete this line.

radhakrishnatg (changes requested):

radhakrishnatg (approved):
Looks good to me.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,890 @@
...
+   "outputs": [],
+   "source": [
+    "# build the flowsheet\n",
+    "m = ConcreteModel()\n",
+    "m.fs = FlowsheetBlock(default={\"dynamic\": False})\n",
+    "\n",
+    "m.fs.thermo_params = GenericParameterBlock(default=thermo_props.config_dict)\n",
+    "\n",
+    "m.fs.M101 = Mixer(default={\"property_package\": m.fs.thermo_params,\n",
+    "                           \"inlet_list\": [\"water_feed\", \"glycol_feed\"]})"
```
dallan-keylogic, line 109: You could take the mixer out and put in Feed/Product blocks if you wanted.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,890 @@
...
+    "permeance_dict = {(0, 'Liq', 'water'): 5611320, (0, 'Liq', 'ethylene_glycol'): 22358.88}  # calculated from literature data\n",
+    "m.fs.pervap.permeance = Var(m.fs.time, m.fs.pervap.phase_list, m.fs.pervap.comp_list,\n",
+    "                            initialize=permeance_dict, units=pyunits.mol/pyunits.s/pyunits.m**2)\n",
+    "m.fs.pervap.permeance.fix()\n",
+    "\n",
+    "m.fs.pervap.area = Var(m.fs.time, initialize=6, units = pyunits.m**2)\n",
+    "m.fs.pervap.area.fix()\n",
+    "\n",
+    "cond_heat_dict = {(0, 'Liq', 'water'): 40.660e3, (0, 'Liq', 'ethylene_glycol'): 56.9e3}\n",
+    "m.fs.pervap.cond_heat = Var(m.fs.time, m.fs.pervap.phase_list, m.fs.pervap.comp_list,\n",
```
dallan-keylogic, line 178: I'm assuming that this is the latent heat of vaporization, but `cond_heat` is pretty cryptic. Ideally we'd use fully readable names.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,890 @@
...
+    }
+   ],
+   "source": [
+    "# separation factor for results analysis - x is inlet water fraction, y is permeate water fraction\n",
+    "m.fs.x = Expression(expr=(m.fs.pervap.inlet.flow_mol_phase_comp[0, \"Liq\", \"water\"] /\n",
+    "                           sum(m.fs.pervap.inlet.flow_mol_phase_comp[0, \"Liq\", i] for i in m.fs.pervap.comp_list)\n",
+    "                          ))\n",
+    "m.fs.y = Expression(expr=(m.fs.pervap.permeate.flow_mol_phase_comp[0, \"Liq\", \"water\"] /\n",
+    "                           sum(m.fs.pervap.permeate.flow_mol_phase_comp[0, \"Liq\", i] for i in m.fs.pervap.comp_list)\n",
+    "                          ))\n",
```
dallan-keylogic, line 611: These are also cryptic names. You could have just named them `inlet_water_frac` and `permeate_water_frac` instead of having a comment to explain your cryptic names

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
```diff
@@ -0,0 +1,890 @@
...
+      "------------------------------------------------------------------------------------\n",
+      "    Stream Table\n",
+      "                                                  Units       water_feed  glycol_feed   Outlet  \n",
+      "    Molar Flowrate ('Liq', 'water')            mole / second     0.34000  1.0000e-06     0.34000\n",
+      "    Molar Flowrate ('Liq', 'ethylene_glycol')  mole / second  1.0000e-06     0.66000     0.66000\n",
+      "    Temperature                                       kelvin      318.15      318.15      318.15\n",
+      "    Pressure                                          pascal  1.0132e+05  1.0132e+05  1.0132e+05\n",
+      "====================================================================================\n",
+      "Question for reviewers: Is there a way to add report() method with a stream table?\n",
+      "\n",
```
dallan-keylogic, line 555: Presently, it doesn't look like there's a way to do this, but it would be good to add, and (probably) not that hard to implement.

Obviously remove this before we merge.

dallan-keylogic (changes requested):
Other than the comments, it would be nice to replace implicit string concatenation with fstrings.

The big thing is getting the "Question for Reviewers" print statement out before merging.

src/Examples/UnitModels/Operations/skeleton_unit_testing.ipynb
bpaul4: As per @radhakrishnatg 's suggestion, I will just remove the comment for now. We can update this example if the `SkeletonModel` gains a `report()` method in the future.

dallan-keylogic (approved):
LGTM

andrewlee94 (approved):
I did a quick skim and it looks good.

PR #137, opened 2022-09-12
(v2) Add setup.py to enable installation as idaes_examples namespace package

### LINKED ISSUE #126: Enable site-packages installation to allow sharing Python modules with common functionality
Needed by #124 (more details to follow...).

bpaul4:
## Summary/Motivation

- This PR adds a `setup.py` file making it possible to install this repository as a "proper" Python package
- The main motivation behind this is being able to access files that are common to multiple examples/notebooks in a robust and consistent manner
- In the current implementation:
  - Packages are located in `pkg` (to minimize conflicts/ambiguity with the existing `src` directory)
  - Only the `hda` and `methanol` Python files have been transferred (this choice was arbitrary and mostly done as a first pass for this draft)

## To install

It should be enough to run `pip install --editable .` from the root of the repository (in addition to the other setup/installation steps). Note that this can conflict with existing examples directories obtained through `idaes get-examples --local` but I'm not sure how exactly this might manifest.

## Open questions/before merging

- [ ] Discuss package structure/import paths, e.g. if we want to use `from idaes_examples.common.hda.flowsheet import ...` or `from idaes_examples.hda.flowsheet import ...` or some other alternativve 
- [ ] Ensure that all the files required by the PRs that need this capability are included here
- [ ] Update the corresponding import paths in the affected notebooks
- [ ] Update CI workflows
- [ ] Update README and other documentation

### Legal Acknowledgement


src/Examples/Flowsheets/methanol_synthesis.ipynb
```diff
@@ -161,7 +161,7 @@
    "metadata": {},
    "source": [
     "## 3.2 Build and Solve Flowsheet\n",
-    "The methanol flowsheet methods are called sequentially below, following the workflow contained in the ```main()``` method in ```idaes/models/flowsheets/methanol_flowsheet.py```. First, let's set the solver options. IDAES contains a default solver `get_solver` which calls IPOPT using standard settings, and we set an iteration cap of 100 to catch nonconverging solver runs."
+    "The methanol flowsheet methods are called sequentially below, following the workflow contained in the ```main()``` method in ```flowsheet.py```. First, let's set the solver options. IDAES contains a default solver `get_solver` which calls IPOPT using standard settings, and we set an iteration cap of 100 to catch nonconverging solver runs."
```
lbianchi-lbl, line 164: I'm wondering what would be the best way to refer to this file now that it lives in a different directory. One method that's a bit roundabout but not likely to break (at least in the short term) would be a link with the full path to the file, pointing to the file on GitHub:

```suggestion
    "The methanol flowsheet methods are called sequentially below, following the workflow contained in the ```main()``` method in [pkg.idaes_examples.common.methanol.flowsheet](https://github.com/IDAES/examples-pse/tree/main/pkg/idaes_examples/common/methanol/flowsheet.py). First, let's set the solver options. IDAES contains a default solver `get_solver` which calls IPOPT using standard settings, and we set an iteration cap of 100 to catch nonconverging solver runs."
```

This is a bit verbose, but at least it makes easier to search and replace when or if we want to use a different scheme.

bpaul4: I like this suggestion, thanks. I'll update the other script references in this manner as well.

Git somehow didn't check-in setup.py (I thought it did) so I'll fix that and the `get_solver` issue to get the tests passing.

pkg/idaes_examples/common/methanol/flowsheet.py
```diff
@@ -44,9 +44,11 @@
 from idaes.models.properties.modular_properties.base.generic_reaction import \
```
andrewlee94, line 44: Do the flowsheets really belong here in the common files, and more generally, are they actually needed/useful at all?

Having the property packages here makes sense as they are used by multiple examples (although even there I think it would be better if they could be stored with the examples that use them), but generally the flowsheet scripts are the examples (basically a copy of the notebook in script form). Thus, I my two questions:

1. As they are basically copies of the examples, shouldn't they go in the folder with the notebooks?
2. As copies, are they really necessary? They effectively duplicate code, doubling the amount of maintenance and increasing the chance of the two versions drifting apart. Originally, I believe we added these to give code-based equivalent to the notebooks in case someone wanted to just run code, but experience suggests that is unlikely.

andrewlee94: Looking more closely at their usage, it seems this is a case where the flowsheet.py file is imported by the notebook rather than duplicating code, so these ones are necessary. I would suggest however that they would be better places in the examples folder.

src/Examples/Flowsheets/methanol_synthesis.ipynb
```diff
@@ -201,6 +201,14 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+      "WARNING: DEPRECATED: The default argument for the ProcessBlock class is\n",
```
andrewlee94, line 204: This appears to have been run with an older version of IDAES - this warning has been fully deprecated now, so we should re-run the notebook using the latest IDAES main.

andrewlee94 (changes requested):
One other comment/request.

pkg/idaes_examples/common/methanol/flowsheet.py
bpaul4: 1. I do agree that flowsheets are less likely to be imported across multiple examples than property packages, and can move the flowsheets back to the main notebook folder if you prefer.
2. Generally, I agree with you that code should not be duplicated. However, in this case the scripts contain long methods and the notebook imports and runs them - the `main()` method essentially does what the notebook does, but the two are not the same and the notebook itself cannot run without either including or importing the methods.

bpaul4: I hadn't seen your second comment - yes I agree with your suggestion and will update the imports as needed.

andrewlee94 (approved):
LGTM

lbianchi-lbl (approved):
As a note to ourselves, this might or might not break the "Run examples" CI workflow in `idaes-pse`, so we should be prepared to fix that if needed (in first approximation, adding `pip install -e .` to whatever commands are used to set up the examples in `.github/actions/run-examples/action.yml`).

PR #158, opened 2022-11-17
Fix NGCC + SOEC cost optimization

eslickj:
## Fixes # .

The cost optimization was failing for the NGCC + SOEC flowsheet.  I found two problems.

 1. There were what a comments said were dummy variables for length and diameter of the SOEC module.  They were initialized to zero but not fixed.  I'm not sure but I think they were there so that the SOEC costing could be build on a generic vessel costing block.  After adding the capital costing equations there were two extra degrees of freedom, which I think were probably the SOEC length and diameter.  I fixed them, but fixing them to 0 failed, so I fixed them to 0.1 m which should be small enough to not matter.  I assume the reason 0 didn't work is that there is a positive lower bound on vessel weight.
 2. The costing constraints and variables were really badly scaled, so I tried to improve the scaling.  I got the condition number of the problem from 3e20 to 2e9.

This fixes the convergence issues, but the result is a bit deferent than the results recored in the SOFC-IES report.  There isn't a huge difference.  I'm not sure if there were changes to the SOEC model or the Costing framework or some other mystery change.  After the report QC, I think we should add some asserts to make sure know if the results change. 

This PR replaces #157.

## Proposed changes:
-

### Legal Acknowledgement

4. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


eslickj:
You can see the changes in the costing file [ngcc_soec_costing.py](https://github.com/IDAES/examples-pse/pull/158#diff-68f2e5c05a5f48c0931dcd8913e766e739470db8f29cacd14cbf56745f4b705c), but the notebook, you can't really see the changes.  I just added a bunch or scaling factors.

bpaul4 (approved):
I'm happy with this, and appreciate your time adding the new scaling. Looking at the notebook output on your branch, both the initial and optimization solves achieve optimality and solve much faster than the current version on main (~100 sec total now compared to 500-800 sec before).

andrewlee94 (approved):
LGTM

PR #163, opened 2022-11-28
Adding Missing Units Models into Jupyter Notebook

MarcusHolly:
## Proposed changes:
Will add information for the following unit models into jupyter notebook

- [x] Feed
- [x] Feed Flash
- [x] Product
- [x] Translator

### Legal Acknowledgement


src/notebook_index.ipynb
```diff
@@ -107,23 +107,24 @@
...
+    "  * [compressor](Examples/UnitModels/Operations/compressor_testing.ipynb) - Compressor Unit Model with Span-Wagner Property Package for supercritical CO2\n",
+    "  * [feed](Examples/UnitModels/Operations/feed_testing.ipynb) - Feed unit model with ideal property package  \n",
+    "  * [heater](Examples/UnitModels/Operations/heater_testing.ipynb) - Heat a liquid mixture of benzene-toluene using a simple heater unit model and an ideal property package\n",
     "\n",
-    "  * [mixer](Examples/UnitModels/Operations/mixer.ipynb) - Mixer unit model with ideal property package\n",
-    "  * [pump](Examples/UnitModels/Operations/pump.ipynb) - Pump unit model with iapws property package\n",
-    "  * [heat exchanger 0D](Examples/UnitModels/Operations/heat%20exchanger%200D.ipynb) - Heat Exchanger 0D unit model heating a benzene-toluene mixture using steam\n",
-    "  * [skeleton unit](Examples/UnitModels/Operations/skeleton%20unit.ipynb) - Skeleton unit model demonstrating custom variables, ports, and constraints\n",
+    "  * [mixer](Examples/UnitModels/Operations/mixer_testing.ipynb) - Mixer unit model with ideal property package\n",
```
andrewlee94, line 114: Just to confirm, are these changes expected?

src/notebook_index.ipynb
```diff
@@ -251,13 +252,33 @@
...
+    "## Contact info\n",
+    "General, background and overview information is available at the [IDAES main website](https://idaes.org).\n",
+    "Framework development happens at our GitHub repo where you can report issues/bugs or make contributions.\n",
+    "For further enquiries, send an email to: idaes-support@idaes.org\n"
    ]
   }
  ],
  "metadata": {
-  "kernel_info": {}
+  "kernel_info": {},
```
andrewlee94, line 263: Similarly, should these changes be here?

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+    "from idaes.core import FlowsheetBlock\n",
+    "\n",
+    "# Import the feed unit model\n",
+    "from idaes.models.unit_models import Feed\n",
+    "\n",
+    "# Import idaes logger to set output levels\n",
+    "import idaes.logger as idaeslog\n",
+    "\n",
+    "# Import the BTX_ideal property package to create a properties block for the flowsheet\n",
+    "from idaes.models.properties.activity_coeff_models import BTX_activity_coeff_VLE\n",
```
andrewlee94, line 75: It would be better to use the modular properties here if possible. The activity coefficient package has not been well maintained, and may need to be changed or deprecated in the future. Plus, it would be good to show the modular properties more.

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The final degrees of freedom is: 0\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Fix the outlet conditions\n",
```
andrewlee94, line 143: 1. This should be feed conditions rather than outlet conditions (as you are fixing things in the "feed" port).
2. I think there needs to be a little be of text explanation before this to indicate that feed blocks have the "feed" port, and what this means.

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The initial degrees of freedom is: 3\n"
+     ]
+    }
+   ],
+   "source": [
```
andrewlee94, line 104: I think there needs to be a brief markdown box before each cell to give a little bit of discussion about the next step.

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+      "\n",
+      "EXIT: Optimal Solution Found.\n",
+      "\b\b\b\b\b\b\b\b\b\b\b\b\b\b"
+     ]
+    }
+   ],
+   "source": [
+    "# Solve the simulation using ipopt\n",
+    "# Note: If the degrees of freedom = 0, we have a square problem\n",
+    "opt = SolverFactory('ipopt')\n",
```
andrewlee94, line 269: We should use the IDAES `get_solver` method instead of the Pyomo solver factory.

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+      "    mole_frac_comp benzene  dimensionless    0.60000\n",
+      "    mole_frac_comp toluene  dimensionless    0.40000\n",
+      "    temperature                    kelvin     298.15\n",
+      "    pressure                       pascal 1.0132e+05\n",
+      "====================================================================================\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Display output report\n",
```
andrewlee94, line 323: A text box before this would be good that says that, as expected, the outlet conditions match the feed conditions we set earlier.

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
...
+     "remove_cell"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "import pytest\n",
+    "\n",
+    "# Check results\n",
+    "assert value(m.fs.feed.outlet.flow_mol[0]) == pytest.approx(100, abs=1e-2)\n",
+    "assert value(m.fs.feed.mole_frac_comp[0, \"benzene\"]) == pytest.approx(0.6, abs=1e-2)\n",
```
andrewlee94, line 342: 1. Relative tolerances are almost always better than absolute tolerances (unless you are dealing with very small numbers).
2. 1e-2 is very loose tolerance, especially given that this is virtually a no-op unit (i.e. we should expect very high tolerance).

andrewlee94 (changes requested):
A few fairly minor requests.

One which I could not attach as a comment: would it be possible to make a better icon for the feed block that is more inline with what you would see on a process flow diagram (i.e. a block arrow pointing right).

src/notebook_index.ipynb
MarcusHolly: Yes because the hyperlinks were not working for some of the existing models because they were missing the "_testing" at the end of the name.

src/notebook_index.ipynb
MarcusHolly: Not sure why/how this got changed. It was not intentional...

src/Examples/UnitModels/Operations/feed_testing.ipynb
```diff
@@ -0,0 +1,376 @@
+{
+ "cells": [
+  {
+   "attachments": {
```
bpaul4, line 4: @MarcusHolly I've encountered a broken HTML link with the "attachments" attribute before, and resolved the issue by using Python Image to load the file instead:

For generic image files, you can write this in Markdown and specify the image size:
""

For SVG files, it may yield better results to write this in Python code:
`from IPython.core.display import SVG
SVG(filename='feed.svg')`

MarcusHolly: Great, thanks Brandon. I'll try this!

src/Examples/UnitModels/Operations/translator_testing.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "# Create the ConcreteModel and the FlowsheetBlock objects, and attach the flowsheet block to it.\n",
+    "m = ConcreteModel()\n",
+    "\n",
+    "m.fs = FlowsheetBlock(dynamic=False) # dynamic or ss flowsheet needs to be specified here\n",
+    "\n",
+    "# Add properties parameter block to the flowsheet with the BT_ideal configuration file\n",
+    "m.fs.properties_ideal = GenericParameterBlock(**configuration_ideal)\n",
+    "\n",
+    "# Add properties parameter block to the flowsheet with the BT_PR configuration file\n",
+    "m.fs.properties_PR = GenericParameterBlock(**configuration_PR)"
```
MarcusHolly, line 106: Still a bit unsure which two property packages I should translate to and from. It seems like all of the modular property packages use the same state variables (FTPx). So in this example, I think the translator block is functioning correctly, it's just not really doing anything since all of the state variables are the same.

andrewlee94: Ah, you need to get an example which uses a different set of state variables. If you look in the tests for the modular properties, you can find some examples using different states - you will need to create a separate configuration dictionary somewhere however (either in the notebook, or as a separate file nearby).

MarcusHolly: Not sure which would be better - putting the configuration dictionary in the notebook or a separate file nearby. As of right now, they are in the notebook. Some of the pros of this are that everything is in one place and showing how the configuration dictionary is set up here could be convenient for a user that wants to create their own. The major con I see is that it may make the translation unit model file seem overly complex.

src/notebook_index.ipynb
lbianchi-lbl: We should look into this before merging. I'm not 100% familiar with the implementation, but I think the `_testing` prefix is stripped by `build.py` during the build process.

My concern is that if the `_testing` prefix is added to the URL, the links will work pre-build (which I assume is why @MarcusHolly had to do that), but not post-build (which is what ultimately is available online). In other words, the links only work pre-build or post-build, but not both.

If this is the case, this is a limitation of the (soon-to-be-retired) `build.py` infrastructure, and I think we should address this (or at least document it) to make it less confusing for example developers.

MarcusHolly:
The images get a bit messed up when converting them from PNG to SVG, so it'd probably be best to replace these whenever the respective unit model icons are added to the IDAES repo.

src/Examples/UnitModels/Operations/feed_flash_testing.ipynb
```diff
@@ -0,0 +1,468 @@
...
+    "\n",
+    "## Learning Outcomes\n",
+    "\n",
+    "- Demonstrate use of the Feed Flash unit model in IDAES\n",
+    "- Demonstrate different options available\n",
+    "\n",
+    "\n",
+    "## Problem Statement\n",
+    "\n",
+    "In this tutorial, we will be feeding vapor-liquid streams into the feed flash unit model using two different sets of state variables. Like the traditional feed block, the feed flash unit model has no inlets and only one outlet, so the stream information is fed to the model through a feed port rather than an inlet. Unlike the feed block, this unit model is capable of handling phase equilibrium. The feed conditions are as follows:\n",
```
andrewlee94, line 19: A little more explanation of when you you use this over the standard Feed port is required to help users understand the difference. Specifically, which sets of state variables does this make sense for, versus which ones do not need this (I think the difference is whether flowrate and composition are specified for the overall mixture (e.g. normal Feed) versus phase or component flows (Feed Flash), but you should confirm that as a it is a long time since I wrote this).

src/Examples/UnitModels/Operations/product_testing.ipynb
```diff
@@ -0,0 +1,291 @@
...
+    "![](product.svg)\n",
+    "\n",
+    "## Learning Outcomes\n",
+    "\n",
+    "- Demonstrate use of the Product unit model in IDAES\n",
+    "\n",
+    "\n",
+    "## Problem Statement\n",
+    "\n",
+    "In this tutorial, we will be feeding a mixed stream of liquid benzene and liquid toluene. This unit model has no outlets and only one inlet. The inlet conditions are as follows:\n",
```
andrewlee94, line 18: Again, an explanation of why you want to include this in a flowsheet is probably required - it is mostly for reporting and visualization purposes as you can build and run a flowsheet without a produce block if you desire.

src/notebook_index.ipynb
```diff
@@ -116,10 +118,12 @@
     "  * [heat exchanger lc](Examples/UnitModels/Operations/heat%20exchanger%20lc.ipynb) - Heat Exchanger Lumped-Capacitance unit model heating a benzene-toluene mixture using steam\n",
     "  * [heat exchanger NTU](Examples/UnitModels/Operations/heat%20exchanger%20NTU.ipynb) - Heat Exchanger NTU unit model for heat integration of two aqueous MEA streams\n",
     "  * [mixer](Examples/UnitModels/Operations/mixer.ipynb) - Mixer unit model with ideal property package\n",
+    "  * [product](Examples/UnitModels/Operations/product.ipynb) - Product unit model with modular property package \n",
     "  * [pump](Examples/UnitModels/Operations/pump.ipynb) - Pump unit model with iapws property package\n",
     "  * [separator](Examples/UnitModels/Operations/separator.ipynb) - Separator unit model with modular property package\n",
-    "  * [skeleton unit](Examples/UnitModels/Operations/skeleton%20unit.ipynb) - Skeleton unit model demonstrating custom variables, ports, and constraints\n",
+    "  * [skeleton unit]Examples/UnitModels/Operations/skeleton%20unit.ipynb) - Skeleton unit model demonstrating custom variables, ports, and constraints\n",
```
andrewlee94, line 124: Looks like you accidentally deleted an opening parenthesis here.

src/Examples/UnitModels/Operations/translator_testing.ipynb
```diff
@@ -0,0 +1,500 @@
...
+    "![](translator.svg)\n",
+    "\n",
+    "## Learning Outcomes\n",
+    "\n",
+    "- Demonstrate use of the Translator unit model in IDAES\n",
+    "\n",
+    "\n",
+    "## Problem Statement\n",
+    "\n",
+    "In this tutorial, we will be transforming an inlet stream that uses the Benzene-Toluene Ideal property package to an outlet stream that uses a user-created configuration file. The state variables for each stream are as follows:\n",
```
andrewlee94, line 18: An explanation of when and why you would want to use a Translator block is important here, as it explains why we need a second property package with different state variables in the example.

andrewlee94 (changes requested):
A few fairly small requests:

1. I would still like to see better icons for the feed and product blocks. A pentagon arrow like the one in Powerpoint would be my starting point.
2. I think there needs to be a little more explanation in the introduction to each of these examples to explain when and why you would want to use each of them. This is especially important for the Feed Flash and Translator blocks, as users will not know what they are and why they would want them.

andrewlee94 (approved):

agarciadiego (approved):
LGTM

PR #166, opened 2022-12-02
Adding StateJunction & Valve Unit Models into Jupyter Notebook

MarcusHolly:
## Proposed changes:
Will add information for the following unit models into jupyter notebook

- [x] StateJunction
- [x] Valve

### Legal Acknowledgement


bpaul4 (commented):
The notebooks look good, but should have a bit more detail describing each step of the examples. I also had a few other comments on the StateJunction example.

src/Examples/UnitModels/Operations/statejunction_testing.ipynb
```diff
@@ -0,0 +1,485 @@
...
+    "# Tutorial: StateJunction Unit Model with Modular Property Package"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from IPython.core.display import SVG\n",
```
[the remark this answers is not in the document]
MarcusHolly: I've tried doing this before and it works locally, but it causes errors on GitHub. If I were to guess what the issue is, it would be because the images I created do not have a transparent background (I've tried doing this myself but couldn't find a way that didn't involve manually doing the tedious work of scrubbing out the background). All of the image files from idaes/models/unit_models/icons are transparent and work fine when printing the image as part of the Markdown.

MarcusHolly:
> The notebooks look good, but should have a bit more detail describing each step of the examples. I also had a few other comments on the StateJunction example.

@bpaul4 Could you specify which steps need more detail? It seems like this is a recurring comment across PRs, so I want to make sure it is adequately addressed. Also, some other examples have very few/no explanations (turbine example). Although the turbine is a much simpler model, I just wasn't sure how much detail should be attached to each step.

bpaul4:
> > The notebooks look good, but should have a bit more detail describing each step of the examples. I also had a few other comments on the StateJunction example.
> 
> @bpaul4 Could you specify which steps need more detail? It seems like this is a recurring comment across PRs, so I want to make sure it is adequately addressed. Also, some other examples have very few/no explanations (turbine example). Although the turbine is a much simpler model, I just wasn't sure how much detail should be attached to each step.

I appreciate the question. The initialization and solve sections are light on text, and a short summary at the end of the example would be useful to loop back to the motivation for the notebook. The examples are sufficient without the text, but it provides additional guidance and clarification for users and is worth adding.

agarciadiego:
> > > The notebooks look good, but should have a bit more detail describing each step of the examples. I also had a few other comments on the StateJunction example.
> > 
> > 
> > @bpaul4 Could you specify which steps need more detail? It seems like this is a recurring comment across PRs, so I want to make sure it is adequately addressed. Also, some other examples have very few/no explanations (turbine example). Although the turbine is a much simpler model, I just wasn't sure how much detail should be attached to each step.
> 
> I appreciate the question. The initialization and solve sections are light on text, and a short summary at the end of the example would be useful to loop back to the motivation for the notebook. The examples are sufficient without the text, but it provides additional guidance and clarification for users and is worth adding.

@MarcusHolly I think you could add a small explanation on why we run the initialize function (i.e., giving all values an initial value) I know it is difficult to understand for first-time users why we run initialize() and then solve(). Additionally, you use WARNING in initialize, you could add a small description to why.

notebook_index.yml
```diff
@@ -115,6 +115,8 @@ contents:
               - pump: Pump unit model with iapws property package
               - heat exchanger 0D: Heat Exchanger 0D unit model heating a benzene-toluene mixture using steam
               - skeleton unit: Skeleton unit model demonstrating custom variables, ports, and constraints
+              - statejunction: Statejunction unit model with modular property package
```
andrewlee94, line 118: I cannot comment on the icons themselves, so I will mention it here: I do not think a box is the best icon for a state junction. It is essentially just a single node (StateBlock) and not an actual unit, so a filled circle might be a better choice.

andrewlee94: Also, do we need to have multiple copies of the icon in different places? It would be nice if we could have a single copy in a central locations (e.g. the `common` folder).

src/notebook_index.ipynb
```diff
@@ -114,6 +114,8 @@
     "  * [pump](Examples/UnitModels/Operations/pump.ipynb) - Pump unit model with iapws property package\n",
     "  * [heat exchanger 0D](Examples/UnitModels/Operations/heat%20exchanger%200D.ipynb) - Heat Exchanger 0D unit model heating a benzene-toluene mixture using steam\n",
     "  * [skeleton unit](Examples/UnitModels/Operations/skeleton%20unit.ipynb) - Skeleton unit model demonstrating custom variables, ports, and constraints\n",
+    "  * [statejunction](Examples/UnitModels/Operations/statejunction.ipynb) - Statejunction unit model with modular property package\n",
```
andrewlee94, line 117: I would remove the "unit model" from this line - as mentioned earlier a `StateJunction` is not a unit model (it is a convenience originally requested by GDP applications).

src/Examples/UnitModels/Operations/statejunction_testing.ipynb
```diff
@@ -0,0 +1,464 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tutorial: StateJunction Unit Model with Modular Property Package\n",
```
andrewlee94, line 7: I would remove "unit model" from here, as with other comments.

src/Examples/UnitModels/Operations/statejunction_testing.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "\n",
+    "## Learning Outcomes\n",
+    "\n",
+    "- Demonstrate use of the StateJunction unit model in IDAES\n",
+    "- Demonstrate different options available\n",
+    "\n",
+    "\n",
+    "## Problem Statement\n",
+    "\n",
+    "In this example, we will be passing two different liquid streams through a StateJunction. The IDAES StateJunction unit model is a pass-through unit or simple pipe that is primarily used to link streams to/from different process alternatives in conceptual design applications. The inlet conditions are as follows:\n",
```
andrewlee94, line 19: Again, I would drop the "unit model" from all of this.

bpaul4 (approved):
Thanks for the quick turnaround, @MarcusHolly, pending the tests I would say this is ready to merge.

src/Examples/UnitModels/Operations/statejunction_testing.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "\n",
+    "**Case 1:**\n",
```
andrewlee94, line 21: A test description of the stream would be nice here (i.e. a mixed stream of benzene and toluene).

src/Examples/UnitModels/Operations/statejunction_testing.ipynb
```diff
@@ -0,0 +1,464 @@
...
+    "from idaes.models.properties.modular_properties.examples.BT_ideal import configuration as configuration\n",
+    "\n",
+    "# Import the saponification property package to create a property block for the case 2 flowsheet\n",
+    "from idaes.models.properties.examples.saponification_thermo import SaponificationParameterBlock\n",
+    "\n",
+    "# Import the degrees_of_freedom function from the idaes.core.util.model_statistics package\n",
+    "# DOF = Number of Model Variables - Number of Model Constraints\n",
+    "from idaes.core.util.model_statistics import degrees_of_freedom\n",
+    "\n",
+    "# Create the ConcreteModel and the FlowsheetBlock objects, and attach the flowsheet block to it.\n",
```
andrewlee94, line 103: It might help to break this into two parts - the imports and the actual flowsheet.

notebook_index.yml
MarcusHolly: For the multiple copies of icons, I agree that we should just have them all in one place. Perhaps this should be part of a separate PR tho since it would involve moving around a bunch of files. So it might just make more sense to make this change after migrating everything over to the new examples repo.

MarcusHolly:
Also, is it worth updating the older examples with better descriptions etc. to bring them up to the standard set by these files? Obviously this is not an immediate issue, but maybe something we (or I) should consider doing at some point down the line.

andrewlee94:
@MarcusHolly Short answer is yes, it would be nice to bring the other examples up to the same standard if we can find the resources.

notebook_index.yml
```diff
@@ -115,7 +115,7 @@ contents:
               - pump: Pump unit model with iapws property package
               - heat exchanger 0D: Heat Exchanger 0D unit model heating a benzene-toluene mixture using steam
               - skeleton unit: Skeleton unit model demonstrating custom variables, ports, and constraints
-              - statejunction: Statejunction unit model with modular property package
+              - statejunction: Statejunction with modular property package
```
andrewlee94, line 118: Typo: Capitalise J in StateJunction, This applies in a few place.

andrewlee94 (commented):
Only one minor comment left - please capitalise the J in StateJunction wherever it appears.

notebook_index.yml
MarcusHolly: I assume the first statejunction in line 118 is fine tho?

agarciadiego (approved):

PR #168, opened 2022-12-06
adding sofc+soec jupyter notebook

AlexNoring:
## Proposed changes:
- Adds an sofc+soec flowsheet from the SOFC-IES project.

### Legal Acknowledgement


bpaul4 (approved):
It may be useful to add unit consistency checks or explicit `pyunits` conversions, especially for the costing, but if the results match the expected values then everything is good. Nicely done!

PR #173, opened 2023-01-19
Update property metadata

andrewlee94:
https://github.com/IDAES/idaes-pse/pull/995 changes how metadata is defined for property packages, and enforces standard naming conventions (or at least forces users to explicitly declare non-standard names). This revealed that some of our examples were using non-standard names, which this PR addresses.

This PR will likely fail without the changes in `idaes-pse`, but this will be necessary to test those changes.

## Proposed changes:
- Update HDA property package to use standard names.
- Updated custom compressor example properties to use more standard names (and drop some unused properties).

### Legal Acknowledgement

### DIFF
--- a/pkg/idaes_examples/common/hda/hda_ideal_VLE.py
+++ b/pkg/idaes_examples/common/hda/hda_ideal_VLE.py
@@ -376,44 +376,35 @@ class HDAParameterData(PhysicalParameterBlock):
     def define_metadata(cls, obj):
         """Define properties supported and units."""
         obj.add_properties(
-            {'flow_mol': {'method': None, 'units': 'mol/s'},
-             'flow_mol_phase_comp': {'method': None, 'units': 'mol/s'},
-             'mole_frac_comp': {'method': None, 'units': 'none'},
-             'temperature': {'method': None, 'units': 'K'},
-             'pressure': {'method': None, 'units': 'Pa'},
-             'flow_mol_phase': {'method': None, 'units': 'mol/s'},
-             'dens_mol_phase': {'method': '_dens_mol_phase',
-                                'units': 'mol/m^3'},
-             'pressure_sat': {'method': '_pressure_sat', 'units': 'Pa'},
-             'mole_frac_phase_comp': {'method': '_mole_frac_phase',
-                                      'units': 'no unit'},
+            {'flow_mol': {'method': None},
+             'flow_mol_phase_comp': {'method': None},
+             'mole_frac_comp': {'method': None},
+             'temperature': {'method': None},
+             'pressure': {'method': None},
+             'flow_mol_phase': {'method': None},
+             'dens_mol_phase': {'method': '_dens_mol_phase'},
+             'pressure_sat': {'method': '_pressure_sat'},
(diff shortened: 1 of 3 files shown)


pkg/idaes_examples/common/hda/hda_ideal_VLE.py
```diff
@@ -376,44 +376,35 @@ def build(self):
...
-             'pressure_bubble': {'method': '_pressure_bubble',
-                                 'units': 'Pa'},
-             'pressure_dew': {'method': '_pressure_dew',
-                              'units': 'Pa'},
-             'fug_vap': {'method': '_fug_vap', 'units': 'Pa'},
-             'fug_liq': {'method': '_fug_liq', 'units': 'Pa'},
-             'dh_vap': {'method': '_dh_vap', 'units': 'J/mol'},
-             'ds_vap': {'method': '_ds_vap', 'units': 'J/mol.K'}})
-
+                     'method': '_enenrgy_internal_mol_phase'},
```
bpaul4, line 391: Typo "enenrgy". I followed the commit history and this typo has been here since the original file was added, which then brings up whether there should have been a warning or error that indicated an incorrect method name.

src/Examples/Advanced/CustomUnitModels/methanol_param_VLE.py
```diff
@@ -107,6 +107,7 @@ def build(self):
              3: ["H2", ("Vap", "Liq")],
              4: ["CH3OH", ("Vap", "Liq")]}
 
+        # Antoine coefficients assume pressure in mmHG and temperature in K
```
bpaul4, line 110: This comment line was duplicated somehow.

bpaul4 (commented):
Just a couple things I noticed, I will re-review after the idaes-pse changes are merged or if the tests pass without it.

pkg/idaes_examples/common/hda/hda_ideal_VLE.py
andrewlee94: @bpaul it would be possible to verify the names, but we generally use the approach and "try it and let it fail then". In this case, it just means this property was never used by an example.

andrewlee94:
@bpaul4 I addressed the issues you found, and the test failures we are seeing as expected (due to the yet-to-be-merged changes in idaes-pse).

However, this PR needs to be merged before the changes in idaes-pse (despite the test failures). We have a chicken-and-egg problem where we have test failures on both sides due to these changes, so one side needs to be merged first with failures, and I think it is best to start here.

bpaul4 (approved):
Approving so the process can move forward.

bpaul4:
FYI it's also possible to fix this PR's head to the corresponding idaes-pse PR branch, rather than merging to test it, if that would be more convenient.

eslickj (approved):

andrewlee94:
@bpaul4 That won't really work - the issue is that neither PR will pass tests until one of them has been merged, so we just have to merge one with failures. Even if we did change the head to test this PR, we would still have to undo the changes before we merge it in which case the failures would be back (and it takes over an hour to run the tests too).

